PR #165, opened 2018-09-01
adds test for user sign in, activates account on sign up.

postables:
# What does this PR fix
* activates user account on sign up (note, this does not enable API access)
# Changes that are being made
* test for sign in added
* account creation function updated to activate user account to allow logins and frontend access.
# Any breaking changes?
* no

### DIFF
--- a/models/user_test.go
+++ b/models/user_test.go
@@ -142,6 +142,48 @@ func TestUserManager_NewAccount(t *testing.T) {
 	}
 }
 
+func TestUserManager_SignIn(t *testing.T) {
+	cfg, err := config.LoadConfig(defaultConfigFile)
+	if err != nil {
+		t.Fatal(err)
+	}
+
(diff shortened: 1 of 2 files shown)


models/user_test.go
```diff
@@ -142,6 +142,48 @@ func TestUserManager_NewAccount(t *testing.T) {
...
+	tests := []struct {
+		name string
+		args args
+	}{
+		{"AccountCreation", args{ethAddress, username, email, "password123", false}},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if _, err := um.NewUserAccount(tt.args.ethAddress, tt.args.userName, tt.args.password, tt.args.email, tt.args.enterpriseEnabled); err != nil {
```
bobheadxi: nice 😁

Issue #170, opened 2018-09-06
Key Generate API response does not include "code" or "response" field on success.


pseudonaut:
**Describe the bug**
Response is:
{status: "key creation sent to backend"}

Should be:
{code: 200, response: "key creation sent to backend"}

It does respond with correct response on invalid token supplied, however:
{code: 401, message: "token contains an invalid number of segments"}

PR #171, opened 2018-09-07
Clean up API response

postables:
Clean up the responses that the API gives
Addresses issues #170 

## :rocket: Changes
Responses now have the format `{"code": <status-code>, "response": <response-from-api>"}`

## :warning: Breaking Changes
No breaking changes to the code itself, but any third-party services would break (at the moment there are none so this okay)

### DIFF
--- a/api/routes_database.go
+++ b/api/routes_database.go
@@ -33,10 +33,7 @@ func (api *API) getUploadsFromDatabase(c *gin.Context) {
 		"service": "api",
 		"user":    authenticatedUser,
 	}).Info("all uploads from database requested")
-	c.JSON(http.StatusOK, gin.H{
-		"code":     http.StatusOK,
-		"response": uploads,
-	})
+	Respond(c, http.StatusOK, gin.H{"response": uploads})
 }
 
 // GetUploadsForAddress is used to read a list of uploads from a particular eth address
@@ -64,8 +61,5 @@ func (api *API) getUploadsForAddress(c *gin.Context) {
 		"user":    user,
 	}).Info("specific uploads from database requested")
 
-	c.JSON(http.StatusOK, gin.H{
-		"code":     http.StatusOK,
-		"response": uploads,
-	})
+	Respond(c, http.StatusOK, gin.H{"response": uploads})
 }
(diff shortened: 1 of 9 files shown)


api/routes_database.go
```diff
@@ -61,5 +64,8 @@ func (api *API) getUploadsForAddress(c *gin.Context) {
 		"user":    user,
 	}).Info("specific uploads from database requested")
 
-	c.JSON(http.StatusFound, gin.H{"uploads": uploads})
+	c.JSON(http.StatusOK, gin.H{
```
bobheadxi: just a thought, but might be cleaner to have a utility function:

```go
package utils
func Respond(c *gin.Context, status int, body gin.H) {
    body["response"] = status
    c.JSON(status, body)
}
```

then:

```go
utils.Respond(c, http.StatusOK, gin.H{"response": uploads})
```

less repetition, and no need to worry about mismatching header/body

bobheadxi:
also, for commit messages, instead of this:

```
fix/api-response: rtfsp response cleanup
```

maybe this would be more descriptive:

```
rtfsp: add status code to API response bodies
```

api/utils.go
```diff
@@ -23,53 +23,48 @@ func CalculateFileSize(c *gin.Context) {
...
 	})
 }
 
 // FailNoExistPostForm is a failure used when a post form does not exist
 func FailNoExistPostForm(c *gin.Context, formName string) {
+	err := fmt.Errorf("%s post form not present", formName)
 	c.JSON(http.StatusBadRequest, gin.H{
-		"error": fmt.Sprintf("%s post form not present", formName),
+		"code":     http.StatusBadRequest,
+		"response": err.Error(),
```
bobheadxi: I dont think this change is needed, isn't this the same as `"response": fmt.Sprintf("%s post form not present", formName)`?

api/utils.go
```diff
@@ -93,3 +88,9 @@ func GetAuthenticatedUserFromContext(c *gin.Context) string {
 	// this is their eth address
 	return claims["id"].(string)
 }
+
+// Respond is a wrapper used to handle API responses
+func Respond(c *gin.Context, status int, body gin.H) {
```
bobheadxi: 🚀

bobheadxi:
Closes #170 i think

PR #174, opened 2018-09-07
IPFS Key Name Change and API Call Check

postables:
Refines IPFS key creation api call
Updates IPFS key names

## :rocket: Changes
Check to see whether or not a user has created a key with the same name before during the API call
Key name has been changed to avoid namespace confliction, they are all prepended with the username
 

## :warning: Breaking Changes
None, just a change to new IPFS key names

### DIFF
--- a/queue/ipfs.go
+++ b/queue/ipfs.go
@@ -90,7 +90,8 @@ func (qm *QueueManager) ProcessIPFSKeyCreation(msgs <-chan amqp.Delivery, db *go
 			d.Ack(false)
 			continue
 		}
-		pk, err := manager.KeystoreManager.CreateAndSaveKey(key.Name, keyTypeInt, bitsInt)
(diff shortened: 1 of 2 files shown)


queue/ipfs.go
```diff
@@ -111,7 +112,7 @@ func (qm *QueueManager) ProcessIPFSKeyCreation(msgs <-chan amqp.Delivery, db *go
 			d.Ack(false)
 			continue
 		}
-		err = userManager.AddIPFSKeyForUser(key.UserName, key.Name, id.Pretty())
+		err = userManager.AddIPFSKeyForUser(key.UserName, keyName, id.Pretty())
```
bobheadxi: as per https://github.com/RTradeLtd/Temporal/wiki/Code-Formatting-Guidelines, this should be inlined right?

postables: yes it should :ok_hand:  good catch

PR #179, opened 2018-09-08
API Content Hash Input Validation

postables:
Validate content hashes being given to us through the API. Previously we would consider anything provided as a hash as valid data. This would get passed further down the line of processing, which is suboptimal. From a security standpoint, this is very dangerous. By validating content hashes, we prevent invalid data from being considered valid

## :rocket: Changes
Anytime an API call is made that provides a content hash as input, we validate it using `go-cid`.

## :warning: Breaking Changes
No, `go-cid` was already a dependent package.

### DIFF
--- a/api/errors.go
+++ b/api/errors.go
@@ -1,0 +1,74 @@
+package api
+
+const (
+	// IPFSConnectionError is an error used for ipfs connection failures
+	IPFSConnectionError = "failed to connect to ipfs"
+	// PrivateNetworkAccessError is used for invalid access to private networks
+	PrivateNetworkAccessError = "invalid access to private netowrk"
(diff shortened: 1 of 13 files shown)


api/errors.go
```diff
@@ -0,0 +1,74 @@
+package api
+
+const (
+	// IPFSConnectionError is an error used for ipfs connection failures
+	IPFSConnectionError = "failed to connect to ipfs"
```
bobheadxi: this shouldn't be in the diff anymore right, now that #177 has been merged? 🤔

postables: Might've been because of how I rebased before-hand? I just did another rebase so it should be good now.

PR #180, opened 2018-09-09
Refined API Get IPFS Keys Call

postables:
Previously when a user would request a list of their IPFS keys, if they had none we would consider it valid and return null data. Now instead, we return an error messaging saying they have no keys

## :rocket: Changes
Adds a new error message, adds a check during the API call which checks the length of key names, and key ids. If either one of them is zero, fail.

## :warning: Breaking Changes
None

### DIFF
--- a/api/errors.go
+++ b/api/errors.go
@@ -71,4 +71,6 @@ const (
 	UserAccountCreationError = "failed to create user account"
 	// PasswordChangeError is an error used when changing your password
 	PasswordChangeError = "failed to change password"
+	// NoKeyError is an error message given to a user when they search for keys, but have none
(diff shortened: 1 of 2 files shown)


api/errors.go
```diff
@@ -71,4 +71,6 @@ const (
 	UserAccountCreationError = "failed to create user account"
 	// PasswordChangeError is an error used when changing your password
 	PasswordChangeError = "failed to change password"
+	// NoKeyError is an error message given to a user when they search for keys, but have none
+	NoKeyError = "no keys"
```
bobheadxi: seems rather brief, though I suppose `no keys found` doesn't say too much more 😅

PR #200, opened 2018-09-25
api: refactors

bobheadxi:

Clean up code before I begin work on TEM-6

## :rocket: Changes

Clean up class properties of API class
Logrus debug level is now set by `debug` argument
Improve logging utility functions and context failing:

- to log and fail a context:

```go
api.LogError(err, KeySearchError)(c, 404)
return
```

- to log with user:

```go
api.LogWithUser(username).Info("ethereum address changed")
```

## :warning: Breaking Changes

Previously, most errors seem to return BadRequest, which seemed inappropriate for the errors being handled - this now defaults to InternalServerError. Code that depends on status codes might break.


bobheadxi:
Coverage dropped because a new package is now being covered, but oh well. I think I'll go ahead and merge it, and we can look into improving coverage in the future

PR #206, opened 2018-09-29
Extract DCCD - Merge Before #2 Release Build

postables:
DCCD was causing some issues with Temporal, and isn't needed to be within the core code base.

## :rocket: Changes
Remove DCCD, put into a new repository

## :warning: Breaking Changes
None


postables:
DCCD is causing a lot of errors, going to merge this through.

PR #214, opened 2018-10-03
API: Add IPNS GET Request

postables:
No way to retrieve the IPNS records a user has published

## :rocket: Changes

* Add a username field to IPNS models
* Add a GET API call

## :warning: Breaking Changes
None


api/routes_ipns.go
```diff
@@ -113,6 +113,18 @@ func (api *API) publishToIPNSDetails(c *gin.Context) {
 	Respond(c, http.StatusOK, gin.H{"response": "ipns entry creation sent to backend"})
 }
 
+// getIPNSRecordsPublishedByUser is used to fetch IPNS records published by a user
+func (api *API) getIPNSRecordsPublishedByUser(c *gin.Context) {
+	username := GetAuthenticatedUserFromContext(c)
+	ipnsManager := models.NewIPNSManager(api.dbm.DB)
```
bobheadxi: similarly to `UserManager`, does this have to be created on a per-API-call basis? I think it can persist through the lifespan of `API`

models/ipns.go
```diff
@@ -99,6 +115,7 @@ func (im *IpnsManager) CreateEntry(ipnsHash, ipfsHash, key, networkName string,
 	entry.TTL = ttl.String()
 	entry.Key = key
 	entry.NetworkName = networkName
+	entry.UserName = username
```
bobheadxi: instead of a bunch of `entry.___`, why not just `entry := IPNS{Sequence: 1, ...}`

models/ipns.go
```diff
@@ -85,7 +100,8 @@ func (im *IpnsManager) UpdateIPNSEntry(ipnsHash, ipfsHash, key, networkName stri
 	return &entry, nil
 }
 
-func (im *IpnsManager) CreateEntry(ipnsHash, ipfsHash, key, networkName string, lifetime, ttl time.Duration) (*IPNS, error) {
+// CreateEntry is used to create a brand new IPNS entry in our database
+func (im *IpnsManager) CreateEntry(ipnsHash, ipfsHash, key, networkName, username string, lifetime, ttl time.Duration) (*IPNS, error) {
```
bobheadxi: not a big deal, but we should be careful of some function signatures getting too large - use of option structs are a good way to manage the complexity of function interfaces

models/ipns.go
postables: good spot!

api/routes_ipns.go
postables: Agreed, perhaps all the model managers should persist through the lifespan of API?

models/ipns.go
postables: :thinking: good idea, that would be especially useful for IPNS record management!

bobheadxi:
I think something went wrong with the merge, and the diff is now kind of wild - how did you do the merge? Looks like the diff from #212 is here now too, but I don't see a corresponding `Merge branch 'V2' into api/ipns-get`

postables:
Yes it does, I will update it :ok_hand:

postables:
Database repo PR to add these changes:

https://github.com/RTradeLtd/database/pull/3

api/api.go
```diff
@@ -35,6 +35,7 @@ type API struct {
 	cfg     *config.TemporalConfig
 	dbm     *database.DatabaseManager
 	um      *models.UserManager
+	im      *models.IpnsManager
```
bobheadxi: 👌

PR #215, opened 2018-10-04
API Cleanup: TEM-75, TEM-76

postables:
There was room for some slight optimizations to the API.

## :rocket: Changes
* On sign up, grant user credits. This is temporary for the beta only
* Removed all ability to remove pins through the API
* Removed API routes to get object size, since that is available through object stat
* HTTP Status Codes given on `FailWithMessage` and `Fail` now default as expected

## :warning: Breaking Changes
* Any applications depending on the removed API calls will not work. Luckily this does not effect the Temporal interface as they were unused routes.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -244,5 +244,5 @@ func (api *API) setupRoutes() {
 	mini.POST("/create/bucket", api.makeBucket)
 	// PROTECTED ROUTES -- END
 
-	api.l.Info("Routes initialized")
+	api.LogInfo("Routes initialized")
 }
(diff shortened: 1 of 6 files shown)


api/api.go
```diff
@@ -202,7 +201,6 @@ func (api *API) setupRoutes() {
 	ipfsPrivateProtected.GET("/network/:name", api.getIPFSPrivateNetworkByName)                          // admin locked
 	ipfsPrivateProtected.POST("/ipfs/check-for-pin/:hash", api.checkLocalNodeForPinForHostedIPFSNetwork) // admin locked
 	ipfsPrivateProtected.POST("/ipfs/object-stat/:key", api.getObjectStatForIpfsForHostedIPFSNetwork)
-	ipfsPrivateProtected.POST("/ipfs/object/size/:key", api.getFileSizeInBytesForObjectForHostedIPFSNetwork)
```
bobheadxi: just a thought but havent really addressed it, shouldn't we have these nested paths (`/ipfs/`) under sub-routers?

postables: good call, this is something we should definitely address before our next release :ok_hand:

PR #216, opened 2018-10-04
Remove Ethereum Addresses From User

postables:
Ethereum addresses as they were  implemented had been deprecated.

## :rocket: Changes
* All references to ethereum addresses on a per-user basis have been removed.
* Hard coded `AdminUser` is now gone
* User model has an `AdminAccess` boolean value

## :warning: Breaking Changes
* Yes, databases will need to be migrated using our migration command.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -27,9 +27,6 @@ import (
 
 var (
 	xssMdlwr xss.XssMw
-
-	// AdminUser is the eth address of the admin account
-	AdminUser = "postables"
 )
 
 // API is our API service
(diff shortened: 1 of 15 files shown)


api/api.go
```diff
@@ -28,8 +28,8 @@ import (
 var (
 	xssMdlwr xss.XssMw
 
-	// AdminAddress is the eth address of the admin account
-	AdminAddress = "0x7E4A2359c745A982a54653128085eAC69E446DE1"
+	// AdminUser is the eth address of the admin account
+	AdminUser = "postables"
```
bobheadxi: this seems incredibly dangerous haha, especially with a public codebase, and we really shouldn't be locking admin functionality behind hardcoded users - at the very least, this should be an environment variable, and this should only be done in some kind of dev mode.

We should update how the users are managed in the database to include data on their permission level, ie admin, and use that instead

postables:
Horrible security practice indeed, I have removed the hard-coded admin user, and switched it to a boolean value in the db.

PR #225, opened 2018-10-13
TEM-90: add utility external systems check route

postables:
Adds a permission-less check used to check basic system status from an external point of view, such as a load balancer, or user wanting to check if the API is online.

## :rocket: Changes
Basic 200 OK utility route which requires 0 permissions.

## :warning: Breaking Changes
None


bobheadxi:
@postables looks like the route got lost in the merge (only the handler is left)

api/api.go
```diff
@@ -161,6 +161,12 @@ func (api *API) setupRoutes() {
 	// V1 API
 	v1 := api.r.Group("/api/v1")
 
+	// external checks used by services like load balances
+	externalChecksProtected := v1.Group("/external-checks")
+	{
+		externalChecksProtected.GET("/systems-check", api.ExternalSystemsCheck)
```
bobheadxi: why not:

```
/external/systems/check
```

this allows better REST expandability:

```
/external/api/check
/external/systems/check
/external/systems/list
```

etc.

bobheadxi: or even better:

```
/status/systems/check
/status/systems/...
/status/...
```

etc

bobheadxi: I feel a lot of the other APIs can be collapsed this way as well, to reduce the number of subroutes and have better grouping, though since that will likely cause compatibility havoc it's not something particularly urgent. I'll open up a ticket tho

api/api.go
```diff
@@ -161,6 +161,12 @@ func (api *API) setupRoutes() {
 	// V1 API
 	v1 := api.r.Group("/api/v1")
 
+	// system checks used to verify the integrity of our services
+	systemChecks := v1.Group("/systems")
```
bobheadxi: just to confirm, these APIs should be accessible to anyone right?

postables: Yup! It's only intended to give a *very* basic response. Ie, status 200 and a basic "systems online" as the response.

PR #229, opened 2018-10-18
Payments - Part 1

postables:
Allows users to purchase credits with ETH, RTC, DASH

## :rocket: Changes
New payment confirmation routes
New payment request routes

## :warning: Breaking Changes
None


api/routes_payment.go
```diff
@@ -1,69 +1,317 @@
...
+	if !exists {
+		FailWithMissingField(c, "tx_hash")
+		return
+	}
+	paymentNumberInt, err := strconv.ParseInt(paymentNumber, 10, 64)
+	if err != nil {
+		Fail(c, err)
+		return
+	}
+	pm := models.NewPaymentManager(api.dbm.DB)
```
bobheadxi: can we initialize `pm` with `api` and hold on to it as a property of `API`?

api/routes_payment.go
```diff
@@ -1,69 +1,317 @@
...
+		paymentNumberString,
+		creditValueFloat,
+		"ethereum",
+		paymentType,
+		username,
+	); err != nil {
+		api.LogError(err, err.Error())(c, http.StatusBadRequest)
+		return
+	}
+	hEncoded := resp.GetH()
```
bobheadxi: what's all this stuff doing? it might be cleaner to move this out of the endpoint - IMO we should try to limit the amount of logic in the endpoints

gapi/eth.go
```diff
@@ -0,0 +1,45 @@
+package gapi
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/RTradeLtd/config"
+	"github.com/RTradeLtd/gapimit/request"
+	"github.com/RTradeLtd/gapimit/response"
+	pb "github.com/RTradeLtd/gapimit/service"
```
bobheadxi: naming on this package seems a bit iffy - calling the repo `github.com/RTradeLtd/temporal-grpc` and calling the package `temporal` might be more useful:

- namespace for package `temporal` is unused
- it makes sense for other services using the gRPC client to call `temporal.NewClient`
- having MIT in the repo name doesnt seem like a good use of name haha

even better, could we just call the repo `RTradeLtd/grpc`? and then the temporal gRPC client could be in in `github.com/RTradeLtd/grpc/temporal`. This allows us to use the repo for other purposes as well:

```
github.com/RTradeLtd/grpc/temporal
github.com/RTradeLtd/grpc/ipfs-orchestrator
github.com/RTradeLtd/grpc/payment
github.com/RTradeLtd/grpc/lens
```

thoughts?

api/routes_payment.go
postables: Yup! I'll just transition to having all "model managers" as properties of `API` ?

gapi/eth.go
postables: Haha I entirely agree, I published that package on git very late at night and couldn't think of something else :joy:

api/routes_payment.go
bobheadxi: sounds good!

bobheadxi:
@postables looks like theres a syntax error being reported by the build

PR #230, opened 2018-10-21
Allow AirDrop Registrations

postables:
Small backend implementation was needed to allow users who receive airdrop codes at devcon to redeem them.

## :rocket: Changes
Temporary DB table to store airdrop information
Temporary API call to redeem airdrop code

## :warning: Breaking Changes
None

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -37,6 +37,7 @@ type API struct {
 	dbm     *database.DatabaseManager
 	um      *models.UserManager
 	im      *models.IpnsManager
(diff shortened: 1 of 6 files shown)


api/api.go
```diff
@@ -37,6 +37,7 @@ type API struct {
 	dbm     *database.DatabaseManager
 	um      *models.UserManager
 	im      *models.IpnsManager
+	dm      *models.DropManager
```
bobheadxi: nice 😁

PR #233, opened 2018-10-23
Payments Part 1 - Again

postables:
Allows users to purchase credits with ETH, RTC, DASH

## :rocket: Changes
New payment confirmation routes
New payment request routes

## :warning: Breaking Changes
None


grpc/eth.go
```diff
@@ -0,0 +1,45 @@
...
+		gconn *grpc.ClientConn
+		err   error
+	)
+	if insecure {
+		gconn, err = grpc.Dial(grpcAPI, grpc.WithInsecure())
+	}
+	if err != nil {
+		return nil, err
+	}
+	sconn := pb.NewSignerClient(gconn)
```
bobheadxi: While this might be an alright start, we probably need to:

- set up TLS
- set up some client-server authentication mechanism

before this goes live

postables: Agreed, tracked in TEM-110

PR #236, opened 2018-10-24
Send Email To Admin On Airdrop Registration

postables:
In preparation for the airdrop distribution at Devcon4, the admin email address for RTrade needs to be notified anytime a user register their airdrop.
This route will be removed before production launch as it is temporary, which is why the hard coded address is okay (imo, anyways)

## :rocket: Changes
Whenever someone confirms their airdrop, send an email to `admin@rtradetechnologies.com`

## :warning: Breaking Changes
None


api/routes_account.go
```diff
@@ -77,6 +77,23 @@ func (api *API) registerAirDrop(c *gin.Context) {
...
+	if err != nil {
+		api.LogError(err, eh.QueueInitializationError)(c, http.StatusBadRequest)
+		return
+	}
+	es := queue.EmailSend{
+		Subject:     "airdrop registered",
+		Content:     fmt.Sprintf("user %s with airdrop code %s to eth address %s is registered", username, aidropID, ethAddress),
+		ContentType: "text/html",
+		UserNames:   []string{"admin"},
+		Emails:      []string{"admin@rtradetechnologies.com"},
```
bobheadxi: not sure how I feel about this hardcoded - probs best to have this be configurable via env, otherwise anyone who tries to spin up their own Temporal will be shooting emails at us haha

bobheadxi: though there'll be a few instances of this email address that needs to be replaced I think

bobheadxi:
```suggestion
		Emails:      []string{os.Getenv("TEMPORAL_ADMIN_EMAIL")},
```

postables: Good idea, i'll create a jira ticket to cleanup hard-coded emails

PR #237, opened 2018-10-25
Advanced Upload Bug Fix: TEM-111

postables:
Due to the way rabbitmq distributes tasks to process advanced file uploads (ie, encrypted uploads) 50% of the requests would land on a worker running on a server, whose corresponding minio instance wasn't storing the file, leading to failures.

## :rocket: Changes
Within the `IPFSFile` message used by rabbitmq, a field for the IP address of the minio host that is storing the file, is included.
Instead of connecting to minio when we start the `IPFSFile` queue,  connect to minio when we find out what instance of minio is hosting the appropriate file.

## :warning: Breaking Changes
None

### DIFF
--- a/api/routes_rtfs.go
+++ b/api/routes_rtfs.go
@@ -170,6 +170,7 @@ func (api *API) addFileLocallyAdvanced(c *gin.Context) {
 	logger.Debugf("file %s stored in minio", objectName)
 
 	ifp := queue.IPFSFile{
+		MinioHostIP:      api.cfg.MINIO.Connection.IP,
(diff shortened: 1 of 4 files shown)


postables:
@bobheadxi Currently with this change we spawn the minio instance during the processing of each message, as opposed to instantiating the connection during the initialization of the queue worker process.

 I'm wondering if it would be better to instantiate the connection to both minio instances during initialization of the queue worker process. Doing this, we could then do a switch on the host ip included in the message to determine which instance we will connect to.

PR #241, opened 2018-11-05
Satisfy Golint: TEM-121

postables:
To actually make the code enterprise grade, golint needs to be satisfied
This will also enable use to trigger hard failures on travis whenever golint isn't satisfied.

## :rocket: Changes
Made changes as pointed out by golint
`enterprise_upload` has been removed as it was a straggler from a very old theoretical implementation of Temporal.

## :warning: Breaking Changes
Hopefully none (however any applications that use the packages within temporal might need to be double checked to ensure they are fine)


.travis.yml
```diff
@@ -20,7 +20,7 @@ script:
   - diff -u <(echo -n) <(gofmt -d -s `find . -type f -name '*.go' -not -path "./vendor/*"`)
   - golint `go list ./... | grep -v /vendor/`
   # TODO: switch on hard fail
```
bobheadxi:
```suggestion
```

postables:
No problem, it had to be done :D

PR #242, opened 2018-11-08
api: add basic lens capabilities

postables:
Depends on https://github.com/RTradeLtd/config/pull/13, so after that gets merged and this gets approved, we will need to update the version dep uses.

Adds basic lens capabilities to the API

## :rocket: Changes
Allow content hashes to be indexed, and searched for

## :warning: Breaking Changes
None


lens/client.go
```diff
@@ -0,0 +1,43 @@
...
+	indexerConn := pb.NewIndexerAPIClient(gconn)
+	return &Client{
+		GC: gconn,
+		IC: indexerConn,
+	}, nil
+}
+
+// SubmitIndexRequest is used to submit an index request to lens
+func (c *Client) SubmitIndexRequest(ctx context.Context, req *pb.IndexRequest) (*pb.IndexResponse, error) {
+	return c.IC.SubmitIndexRequest(ctx, req)
```
bobheadxi: if the member IC is public, theres no need to wrap its API right? you can just go `c.IC.....` externally, or even better embed the class:

```go
type Client struct {
	GC *grpc.ClientConn
	pb.IndexerAPIClient
}
```

or make `ic` private

lens/client.go
```diff
@@ -0,0 +1,43 @@
...
+	if insecure {
+		gconn, err = grpc.Dial(url, grpc.WithInsecure())
+	}
+	if err != nil {
+		return nil, err
+	}
+	indexerConn := pb.NewIndexerAPIClient(gconn)
+	return &Client{
+		GC: gconn,
+		IC: indexerConn,
```
bobheadxi: this will null ptr very badly if `insecure` is false

on the `orchestrator` branch of the grpc repo I've added utilities for attaching credentials to requests https://github.com/RTradeLtd/grpc/blob/orchestrator/utils/credentials.go, there are examples to using this in the ipfs-orchestrator for now

i guess its fine for now though

postables: Nope it should be done right away, it removes the chance of forgetting to  do so before we launch. I'll message you on slack when this PR is ready to go.

postables:
TODO: make search case-insensitive

api/routes_lens.go
```diff
@@ -68,5 +68,9 @@ func (api *API) submitSearchRequest(c *gin.Context) {
 		api.LogError(err, eh.FailedToSearchError)(c, http.StatusBadRequest)
 		return
 	}
+	if len(resp.Names) == 0 {
+		api.LogError(err, eh.NoSearchResultsError)(c, http.StatusBadRequest)
```
bobheadxi: I don't think this qualifies as an error in terms of logging - at least how I see it is when we log an error, its something kinda to very critical, that needs attention, whereas someone searching "asdfjalsef" and turning up no results isn't something we need to be immediately aware of

bobheadxi: we might want to do an info-level log maybe, since it might be useful for watching search trends

lens/client.go
```diff
@@ -0,0 +1,51 @@
...
+	defaultURL = "127.0.0.1:9998"
+)
+
+// Client is a lens client used to make requests to the Lens gRPC server
+type Client struct {
+	pb.IndexerAPIClient
+}
+
+// NewClient is used to generate our lens client
+func NewClient(opts config.Endpoints) (*Client, error) {
```
bobheadxi: i noticed that the signer client is in the `grpc` folder here, it might be nice to have all the grpc clients used by temporal inside `grpc`? or maybe rename it to `grpc_clients` to avoid conflicts with the grpc repo

api/routes_lens.go
postables: Good call

PR #250, opened 2018-11-13
Fix Private Network Integration

postables:
Iron out a few small bugs
Add remaining endpoints

## :rocket: Changes
Allow for start/stop of networks
Fix bug with network creation attempting to add permissions for the logged in user twice
Basic stats endpoint (this will need to be fine-tuned and allow for users to select which stats to receive)

## :warning: Breaking Changes
Private IPFS network api paths changed slightly


api/routes_rtfsp.go
```diff
@@ -567,7 +567,8 @@ func (api *API) createHostedIPFSNetworkEntryInDatabase(c *gin.Context) {
 		return
 	}
 	logger.WithField("user", username).Info("network added to user")
-	if len(users) > 0 {
+	// if there is only 1 user, this next part is covered by the previous logic
+	if len(users) > 1 {
```
bobheadxi: `users` always includes the user (https://github.com/RTradeLtd/Temporal/blob/acc6b41ef7eb70009b1ae714f64c7992391296da/api/routes_rtfsp.go#L550) - if anything, the previous logic should be removed in favour of the user iteration

api/routes_rtfsp.go
```diff
@@ -641,6 +686,50 @@ func (api *API) stopIPFSPrivateNetwork(c *gin.Context) {
...
+	if err != nil {
+		api.LogError(err, "failed to get network stats")(c, http.StatusBadRequest)
+		return
+	}
+	Respond(c, http.StatusOK, gin.H{
+		"response": gin.H{
+			"network_name": networkName,
+			"disk_usage":   stats.GetDiskUsage(),
+			"uptime":       stats.GetUptime(),
+			"api":          api.cfg.Orchestrator.Host + stats.GetApi(),
```
bobheadxi: this doesnt seem right... the host is set by the orchestrator: https://github.com/RTradeLtd/ipfs-orchestrator/blob/master/orchestrator/orchestrator.go#L244

bobheadxi: also i've already integrated stats reporting here: https://github.com/RTradeLtd/Temporal/blob/V2/api/routes_rtfsp.go#L649

i think we should keep just one of the two

postables: Hmm I wonder if it's an error then. Everytime the API would be returned as `:<port-number>` such as `:5432` unless that is intended? We'll keep the stats report you already created.

api/routes_rtfsp.go
postables: yes the previous logic should be removed

api/routes_rtfsp.go
bobheadxi: I think you might have forgotten to set the address in the orchestrator config - try generating a new one, its a fairly recent addition

.travis.yml
```diff
@@ -10,6 +10,8 @@ install:
   - go get -u golang.org/x/lint/golint
 
 before_script:
+  - sudo mkdir -p /ipfs/keystore
+  - sudo chmod -R 777 /ipfs
```
bobheadxi: is this path fixed? can this be configured via a configuration option so that a non-root path can be used?

api/routes_rtfsp.go
```diff
@@ -600,6 +597,49 @@ func (api *API) createHostedIPFSNetworkEntryInDatabase(c *gin.Context) {
...
+	networkName, exists := c.GetPostForm("network_name")
+	if !exists {
+		FailWithMissingField(c, "network_name")
+		return
+	}
+	logger := api.LogWithUser(username).WithField("network_name", networkName)
+	logger.Info("private ipfs network start requested")
+
+	// verify access to the requested network
+	networks, err := api.um.GetPrivateIPFSNetworksForUser(username)
```
bobheadxi: we might want a helper function to do this check, I can put in a PR later one this lands

grpc-clients/orchestrator.go
```diff
@@ -31,7 +31,10 @@ func NewOcrhestratorClient(opts config.Orchestrator, devMode bool) (*IPFSOrchest
 			grpc.WithTransportCredentials(creds),
 			grpc.WithPerRPCCredentials(dialer.NewCredentials(opts.Key, !devMode)))
 	} else {
-		dialOpts = append(dialOpts, grpc.WithInsecure())
+		dialOpts = append(
+			dialOpts,
```
bobheadxi: dialOpts should be on the first line I think, since its not the same argument type as the rest (which are part of the spread argument)

```go
dialOpts = append(dialOpts,
    /* other stuff */,
    /* other stuff */)
```

api/routes_rtfsp.go
postables: Sounds good

.travis.yml
postables: Yup, it's hard-coded within the temporal code-base.

bobheadxi: actually it looks like it can be configured already: https://sourcegraph.com/github.com/RTradeLtd/Temporal@dc4e58eb9050ff2897a13493a2ec0c198a599a99/-/blob/rtfs/keystore.go#L25

actually, based on the [references](https://sourcegraph.com/github.com/RTradeLtd/Temporal@dc4e58eb9050ff2897a13493a2ec0c198a599a99/-/blob/rtfs/keystore.go#L25:6&tab=references), it looks like no tests are actually using the default path?

api/routes_rtfsp.go
```diff
@@ -641,6 +681,50 @@ func (api *API) stopIPFSPrivateNetwork(c *gin.Context) {
...
+	if !found {
+		logger.Info("user not authorized to access network")
+		Respond(c, http.StatusUnauthorized, gin.H{
+			"response": "user does not have access to requested network",
+		})
+		return
+	}
+	if _, err = api.orch.RemoveNetwork(c, &ipfs_orchestrator.NetworkRequest{
+		Network: networkName}); err != nil {
+		api.LogError(err, "failed to stop network")(c)
```
bobheadxi: this should be `failed to remove network assets` or something along those lines

api/routes_rtfsp.go
bobheadxi: I've added a warning to the orchestrator if no address is set: https://github.com/RTradeLtd/ipfs-orchestrator/commit/7e8e0a84ea070ada3ecf6d2bad18adaa4909ee59

api/routes_rtfsp.go
```diff
@@ -561,12 +561,8 @@ func (api *API) createHostedIPFSNetworkEntryInDatabase(c *gin.Context) {
...
 	logger.WithField("db_id", network.ID).Info("database entry created")
 
-	// add network to users
-	if err := api.um.AddIPFSNetworkForUser(username, networkName); err != nil {
-		api.LogError(err, eh.NetworkCreationError)(c)
-		return
-	}
 	logger.WithField("user", username).Info("network added to user")
+	// if there is only 1 user, this next part is covered by the previous logic
 	if len(users) > 0 {
```
bobheadxi: `len(users)` should always be `> 0`: https://github.com/RTradeLtd/Temporal/blob/0717632842fbafa485044e7eb71232b4be01b5c9/api/routes_rtfsp.go#L548

also I dont think this comment is valid anymore

the log on 564 isnt needed anymore either i think

.travis.yml
postables: Line 130 in `api.go` initializes the keystore manager with its IPFS connection, which uses a default path of `/ipfs/keystore` in `rtfs/utils.go`.

api/routes_rtfsp.go
postables: Good catch

.travis.yml
bobheadxi: We should make that part of API configuration then 👍

.travis.yml
```diff
@@ -10,6 +10,8 @@ install:
   - go get -u golang.org/x/lint/golint
 
 before_script:
+#  - sudo mkdir -p /ipfs/keystore
+#  - sudo chmod -R 777 /ipfs
```
bobheadxi: we can probably remove these commented lines

rtfs/keystore.go
```diff
@@ -27,7 +27,7 @@ func GenerateKeystoreManager(keystorePath ...string) (*KeystoreManager, error) {
 		storePath = DefaultFSKeystorePath
 		km        KeystoreManager
 	)
-	if keystorePath != nil && len(keystorePath) > 0 {
+	if keystorePath != nil && len(keystorePath) > 0 && keystorePath[0] != "" {
```
bobheadxi: 👍

PR #258, opened 2018-11-24
Update RTFS package, cleanup API calls

postables:
Wait until https://github.com/RTradeLtd/rtfs/pull/7 is merged, then create a new release for `rtfs` and update the version used by this PR, before merging.

Needed to update the `rtfs` version we use
Cleaned up API calls
Better handling of auth tokens
Small amount of additional security measures
[Postman documentation has been updated](https://documenter.getpostman.com/view/4295780/RWEcQM6W)
## :rocket: Changes
`rtfs` was updated to include `krab` and `beam`

API calls were a little messy for IPFS usage, and they have been adjusted

Currently whenever we go to retrieve usernames from auth tokens, it can be possible for empty auth tokens to be provided. This is catched by our other protections, however by not checking for this, we potentially open up security holes. If we can't recover a username from the auth token, then we abort processing their request and return an error message

Small amount of additional security measures were added to the API, mostly related to XSS filtering, and cache control.

## :warning: Breaking Changes
Yes, basically every IPFS api call is renamed


api/api.go
```diff
@@ -209,12 +216,34 @@ func (api *API) ListenAndServe(addr string, tls *TLSConfig) error {
...
+			log.Fatal(err)
+		}
+	}
 	// set up defaults
 	api.r.Use(
+		// slightly more complex xss removal
 		xssMdlwr.RemoveXss(),
-		limit.MaxAllowed(20),
+		// rate limiting
+		limit.MaxAllowed(connLimit),
```
bobheadxi: 👍 configuration injection!

api/routes_mini.go
```diff
@@ -11,7 +11,11 @@ import (
 
 // MakeBucket is used to create a bucket in our minio container
 func (api *API) makeBucket(c *gin.Context) {
-	username := GetAuthenticatedUserFromContext(c)
+	username, err := GetAuthenticatedUserFromContext(c)
+	if err != nil {
+		api.LogError(err, eh.NoAPITokenError)(c, http.StatusBadRequest)
```
bobheadxi: shouldn't JWT middleware prevent unauthenticated API requests from getting through?

api/routes_rtfsp.go
```diff
@@ -779,16 +828,99 @@ func (api *API) getUploadsByNetworkName(c *gin.Context) {
...
+		api.LogError(err, eh.NoAPITokenError)(c, http.StatusBadRequest)
+		return
+	}
+	forms := api.extractPostForms(c, "source_network", "destination_network", "content_hash")
+	if len(forms) == 0 {
+		return
+	}
+	var (
+		net1URL string
+		net2URL string
```
bobheadxi: why not `source` and `dest`?

bobheadxi: I think you can also declare this as:

```go
var source, dest string
```

api/routes_mini.go
postables: That's a good point :ok_hand:

api/api.go
```diff
@@ -209,12 +216,34 @@ func (api *API) ListenAndServe(addr string, tls *TLSConfig) error {
...
+		connLimit int
+		err       error
+	)
+	if api.cfg.API.Connection.Limit == "" {
+		connLimit = 50
+	} else {
+		// setup the connection limit
+		connLimit, err = strconv.Atoi(api.cfg.API.Connection.Limit)
+		if err != nil {
+			log.Fatal(err)
```
bobheadxi: we might want to have this function return err instead of panicking

api/routes_utils.go
```diff
@@ -1,12 +1,95 @@
...
+
+// BeamContent is used to beam content from one network to another
+func (api *API) beamContent(c *gin.Context) {
+	username, err := GetAuthenticatedUserFromContext(c)
+	if err != nil {
+		api.LogError(err, eh.NoAPITokenError)(c, http.StatusBadRequest)
+		return
+	}
+	forms := api.extractPostForms(c, "source_network", "destination_network", "content_hash")
+	if len(forms) == 0 {
```
bobheadxi: this returns `nil` if something is missing - `len(forms)` will panic

bobheadxi: oops i stand corrected

PR #260, opened 2018-11-28
docs: add subrepos to README, switch to Apache-2.0

bobheadxi:

We have quite a lot of subrepos - they're (I think) currently all tagged with `temporal`, so a link to a github filter can bring up all related subrepos.

All other temporal components are under apache-2.0 now, so this PR also switches to the new licence

## :rocket: Changes

link to subrepos

mit -> apache2.0

## :warning: Breaking Changes

might break fossa

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,21 +1,201 @@
-MIT License
+                                 Apache License
+                           Version 2.0, January 2004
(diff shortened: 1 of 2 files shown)


bobheadxi:
oh oops forgot to push version bump, will open another PR

PR #262, opened 2018-11-29
Allow for the creation of multiple workers for a single queue

postables:
Currently we were limited to a single worker per queue per node, which limits the total amount of requests we can process at a single moment. This allows us to create any number of workers per queue on a single node, defaulting to 2. 

This is the first of two changes to the queue system. The next change will focus on optimizing the code.

## :rocket: Changes
Spawn multiple workers in go-routines, allowing graceful shutdowns of the channels, and connections we generate between rabbitmq

## :warning: Breaking Changes
None


queue/queue.go
```diff
@@ -136,20 +142,26 @@ func (qm *Manager) DeclareQueue() error {
 // ConsumeMessage is used to consume messages that are sent to the queue
 // Question, do we really want to ack messages that fail to be processed?
 // Perhaps the error was temporary, and we allow it to be retried?
-func (qm *Manager) ConsumeMessage(consumer, dbPass, dbURL, dbUser string, cfg *config.TemporalConfig) error {
+func (qm *Manager) ConsumeMessage(ctx context.Context, wg *sync.WaitGroup, consumer, dbPass, dbURL, dbUser string, cfg *config.TemporalConfig) error {
```
```suggestion
func (qm *Manager) ConsumeMessages(ctx context.Context, wg *sync.WaitGroup, consumer, dbPass, dbURL, dbUser string, cfg *config.TemporalConfig) error {
```

postables:
reminder to self: remove sslmodedisable before moving PR to review status

PR #264, opened 2018-12-01
Queue Cleanup + Additional Tests: TEM-126

postables:
Beef up our processing capabilities tremendously, and reduce API response time. Theoretically, this change should allow us to process 10x the amount of messages on a given node, and reduce API response time (and resource usage) for certain calls.

During testing, API calls interacting with the queue system appear to be 2-3ms faster on average

Enabled additional tests in varying capacities. After this PR is merged, I'll start integrating full tests within the API.

## :rocket: Changes
* Anytime a message is received on a queue, it's pulled out and processed in a goroutine, allowing for additional messages to be pulled out and processed simultaneously. 
* Instead of hard limiting the number of goroutines that can be launched, we apply a QoS setting, allowing for at most 10 unacknowledged messages on a given queue. This means that at most, 10 messages can be processed at the same time 
* Allow graceful shutdown of queue service/consumer connections
* Allow graceful shutdown of queue publisher connections
* Instead of initializing a connection to a queue every single time an API call is made which interacts with the queue, connections are instantiated during api initialization.
* Allow private keys creation for private networks. Since we use `krab` we can share private keys across multiple networks.

## :warning: Breaking Changes
None


api/api.go
```diff
@@ -185,17 +216,35 @@ func new(cfg *config.TemporalConfig, router *gin.Engine, ipfs rtfs.Manager, debu
...
 	api.signer.Close()
 	api.orch.Close()
+	// close queue resources
+	api.queues.cluster.Close()
+	api.queues.database.Close()
+	api.queues.email.Close()
+	api.queues.file.Close()
+	api.queues.ipns.Close()
+	api.queues.key.Close()
+	api.queues.pin.Close()
```
bobheadxi: weeee probs no way to do this better but damn thats a lotta boilerplate and things we might forget to `Close()` lmao

queue/ipfs_cluster.go
```diff
@@ -1,81 +1,84 @@
...
-			}
-		} else {
-			_, err = uploadManager.UpdateUpload(clusterAdd.HoldTimeInMonths, clusterAdd.UserName, clusterAdd.CID, clusterAdd.NetworkName)
-			if err != nil {
-				qm.LogError(err, "failed to update upload in database")
+	for {
+		select {
+		case d := <-msgs:
+			wg.Add(1)
+			go func(d amqp.Delivery) {
```
bobheadxi: this could probably be a separate function, this is getting pretty indented

as a separate function would make this more testable too! 😁

queue/ipns.go
```diff
@@ -38,95 +37,87 @@ func (qm *Manager) ProcessIPNSEntryCreationRequests(msgs <-chan amqp.Delivery, d
...
-				ie.UserName,
-				ie.LifeTime,
-				ie.TTL,
-			); err != nil {
-				qm.LogError(err, "failed to create ipns entry in database")
+	for {
+		select {
+		case d := <-msgs:
+			wg.Add(1)
+			go func(d amqp.Delivery) {
```
bobheadxi: same here

cmd/temporal/main.go
```diff
@@ -120,10 +154,20 @@ var commands = map[string]cmd.Cmd{
...
-							if err != nil {
-								log.Fatal(err)
-							}
+							quitChannel := make(chan os.Signal)
+							signal.Notify(quitChannel, os.Interrupt, syscall.SIGTERM, syscall.SIGINT)
+							ctx, cancel := context.WithCancel(context.Background())
+							waitGroup := &sync.WaitGroup{}
+							waitGroup.Add(1)
+							go func() {
+								if err := qm.ConsumeMessages(ctx, waitGroup, "", args["dbPass"], args["dbURL"], args["dbUser"], &cfg); err != nil {
```
bobheadxi: hm seeing a few of these in the cmd code, why does this have to run in a goroutine?

queue/queue.go
```diff
@@ -13,81 +15,56 @@ import (
...
-	}
-	qm.QueueName = fmt.Sprintf("%s+%s", host, queueName)
+	qm.logger = logger
+	qm.logger.Info("Logging initialized")
 	return nil
 }
 
 // Initialize is used to connect to the given queue, for publishing or consuming purposes
-func Initialize(queueName, connectionURL string, publish, service bool) (*Manager, error) {
+func Initialize(queueName, connectionURL string, publish, service bool, logFilePath ...string) (*Manager, error) {
```
```suggestion
func New(queue, URL string, publish, service bool, logFilePath ...string) (*Manager, error) {
```

also, why is `service` a `bool`? it looks like it enables logging

queue/queue.go
```diff
@@ -13,81 +15,56 @@ import (
...
 	}
-	qm := Manager{Connection: conn}
+	qm := Manager{connection: conn}
+	// open a channel
 	if err := qm.OpenChannel(); err != nil {
 		return nil, err
 	}
-
+	// setup queue nname
 	qm.QueueName = queueName
```
bobheadxi: why not just put this when you instantiate `qm`?

queue/queue.go
```diff
@@ -13,81 +15,56 @@ import (
...
-func Initialize(queueName, connectionURL string, publish, service bool) (*Manager, error) {
+func Initialize(queueName, connectionURL string, publish, service bool, logFilePath ...string) (*Manager, error) {
 	conn, err := setupConnection(connectionURL)
 	if err != nil {
 		return nil, err
 	}
-	qm := Manager{Connection: conn}
+	qm := Manager{connection: conn}
+	// open a channel
 	if err := qm.OpenChannel(); err != nil {
```
bobheadxi: `OpenChannel` is only ever called here, this should be private

bobheadxi: I think the function should be done inline actually - it might be best to instantiate as much of `qm` as possible in one go, instead of doing it piecewise - one thing I like to do is gather all dependencies, and as late as possible, create the relevant class:

```go
return &Manager{ /* a bunch of stuff */ }, nil
```

queue/queue.go
```diff
@@ -126,31 +103,40 @@ func (qm *Manager) DeclareQueue() error {
...
 	return nil
 }
 
-// ConsumeMessage is used to consume messages that are sent to the queue
+// ConsumeMessages is used to consume messages that are sent to the queue
 // Question, do we really want to ack messages that fail to be processed?
 // Perhaps the error was temporary, and we allow it to be retried?
-func (qm *Manager) ConsumeMessage(consumer, dbPass, dbURL, dbUser string, cfg *config.TemporalConfig) error {
+func (qm *Manager) ConsumeMessages(ctx context.Context, wg *sync.WaitGroup, consumer string, sslModeDisable bool, cfg *config.TemporalConfig) error {
 	db, err := database.OpenDBConnection(database.DBOptions{
```
bobheadxi: shouldnt this db and stuff be part of the main constructor, or are there instances of the queue where you dont need the DB?

bobheadxi: if the latter, you might want to pass around the `db` instead of making the various message consumption helpers hope that `qm` is initialized with a database manager, which makes the dependency more explicit. For example, init the DB here, and then:

```go
	case DatabaseFileAddQueue:
		return qm.ProcessDatabaseFileAdds(ctx, wg, db, msgs)
```

even better, since this will probably be initialized alongside a bunch of other stuff, perhaps `ConsumeMessages` can accept a database as a parameter as well? So as to create the connection once, and pass it off to classes that depend on it

bobheadxi: imo it's good practice to streamline dependency trickle-downs, and make sure we know who needs what, with the idea that each class should need as little as possible

cmd/temporal/main.go
```diff
@@ -368,24 +369,27 @@ func main() {
...
-	sslModeDisable := os.Getenv("SSL_MODE_DISABLE")
-	sslDisabled := false
-	if sslModeDisable != "" {
-		sslDisabled, err = strconv.ParseBool(sslModeDisable)
+	initDB := os.Getenv("INIT_DB")
+	if strings.ToLower(initDB) == "yes" {
+		sslModeDisable := os.Getenv("SSL_MODE_DISABLE")
+		sslDisabled := false
+		if sslModeDisable != "" {
+			sslDisabled, err = strconv.ParseBool(sslModeDisable)
```
bobheadxi: this whole chunk could be:

```go
var sslDisabled = os.Getenv("SSL_MODE_DISABLE") == "true"
```

queue/queue.go
postables: I was able to re-organize the order of calls for that function, however `Manager` can only be instantiated with the string variables, the rest of the setup requires invocation in a particular order.

queue/queue.go
postables: Changed it

api/api.go
postables: Haha yea can't think of a better way :frowning_face:  I added some logging so we can tell after the fact whether or not the channels were properly closed

api/api.go
```diff
@@ -185,17 +228,51 @@ func new(cfg *config.TemporalConfig, router *gin.Engine, ipfs rtfs.Manager, debu
 		signer:  signer,
 		orch:    orch,
 		dc:      dc,
-		zm:      models.NewZoneManager(dbm.DB),
-		rm:      models.NewRecordManager(dbm.DB),
-		nm:      models.NewHostedIPFSNetworkManager(dbm.DB),
+		queues: queues{
```
bobheadxi: nice 👍

queue/queue.go
```diff
@@ -126,31 +99,30 @@ func (qm *Manager) DeclareQueue() error {
...
-		Port:     cfg.Database.Port,
-	})
-	if err != nil {
-		return err
-	}
-
-	// ifs the queue is using an exchange, we will need to bind the queue to the exchange
+func (qm *Manager) ConsumeMessages(ctx context.Context, wg *sync.WaitGroup, db *gorm.DB, cfg *config.TemporalConfig) error {
+	// embed database into queue manager
+	qm.db = db
```
bobheadxi: hm, i'd prefer the manager to not have a db and have the db be passed to each processor as needed, but this is fine for now i think

setup/scripts/temporal_manager.sh
```diff
@@ -13,33 +13,43 @@ export CONFIG_DAG
 case "$1" in
 
     api)
+        INIT_DB=yes
+        export INIT_DB
         temporal api
         ;;
     queue-dfa)
+        INIT_DB=yes
```
bobheadxi: hm why not `true` instead of `yes` (since `true` as env toggles is already used elsewhere in the codebase)

postables: Done

queue/queue.go
postables: I'm not sure what the best choice is. It's pretty much used by every single consumer for a few purposes:
1) issuing refunds
2) instantiating our model managers.

In my eyes, it's a property of the manager, so it seems like it would make sense to embed it within the manager object.

Thoughts?

If we decide it is best like that, I'll do that during my next pass-through of the queue system.

postables:
Thanks :fire:  Got a lot more coverage increases coming, will mock out a complete (or near complete) API test suite next.

queue/queue.go
bobheadxi: You noted that only consumers use it, nor publishers, which is where the main issue is I guess - perhaps they shouldn’t be tied together under one class?

PR #265, opened 2018-12-04
api: separate v2 api into its own package, v3 prototype

bobheadxi:

Just for clarity, and gives us room to start thinking about V3.

Calling it V2 after brief discussion with @postables about the state of v1 semver (nonexistent 😋) and the possibility of launching with V2 to give us a blank slate to work with. Also the branch is called v2 already

### V2

change base path of all URLs to `/v2` instead of `/v1`
rename package `api` to `v2`
change import path from `Temporal/api` to `Temporal/api/v2`

### V3

Prototype v3 stuff, doesnt really work/do anything yet though

## :warning: Breaking Changes

API paths now `/v2/...`


api/v2/api.go
```diff
@@ -1,6 +1,6 @@
-// Package api is the main package for Temporal's
+// package v2 is the main package for Temporal's
```
postables:
```suggestion
// Package v2 is the main package for Temporal's
```

PR #270, opened 2018-12-11
api: adjust network removal logic order + TEM-167

postables:
Network removal was giving errors due to the order of `remove network -> search for network -> remove network from authorized users`

Start mocking out basic tests of the gRPC  (lens+orchestrator) services

Some test refactor, along with adding a test for the api call within our `cmd` package.

Remove duplicated code for content hash downloading

Fixed password change issue in TEM-167 which was due to htmlencoding of string. 

TODO: Update frontend code to reflect the new download link

## :rocket: Changes
Network removal logic was changed to `search for network -> remove network -> remove network from authorized users`

Add basic tests for ipfs-orchestrator, and lens api calls

The download content hash functionality has been reworked to reduce a lot of duplicated code. Instead of a download function specific to public, or private IPFS. We now have generic download function, which allows the user to optionally specify the network to download from. If the user doesn't specify, then it is considered public.
## :warning: Breaking Changes
None


api/v2/api_test.go
```diff
@@ -37,15 +37,16 @@ const (
...
+	hash             = "QmPY5iMFjNZKxRbUZZC85wXb9CFgNSyzAy1LxwL62D8VGr"
+	api              *API
+	db               *gorm.DB
+	engine           *gin.Engine
+	testRecorder     *httptest.ResponseRecorder
+	authHeader       string
+	cfg              *config.TemporalConfig
+	fakeLensClient   *mocks.FakeIndexerAPIClient
+	fakeOrchClient   *mocks.FakeServiceClient
+	fakeSignerClient *mocks.FakeSignerClient
```
bobheadxi: hm dont think we should do this, mocks should be created on a per-test basis because theres counting stuff that goes on under the hood

api/v2/api.go
```diff
@@ -285,11 +278,24 @@ type TLSConfig struct {
...
 
 // ListenAndServe spins up the API server
-func (api *API) ListenAndServe(addr string, tls *TLSConfig) error {
-	if tls != nil {
-		return api.r.RunTLS(addr, tls.CertFile, tls.KeyFile)
+func (api *API) ListenAndServe(ctx context.Context, addr string, tls *TLSConfig) error {
+	errChan := make(chan error, 1)
+	go func() {
+		if tls != nil {
+			errChan <- api.r.RunTLS(addr, tls.CertFile, tls.KeyFile)
```
bobheadxi: i dont think this is needed, just return, and have `RunTLS` in main thread

api/v2/api.go
```diff
@@ -285,11 +278,24 @@ type TLSConfig struct {
...
+		}
+		errChan <- api.r.Run(addr)
+		return
+	}()
+	for {
+		select {
+		case err := <-errChan:
+			return err
+		case <-ctx.Done():
+			return nil
```
bobheadxi: run this select in a goroutine instead, do `api.r.Close` when ctx finishes

bobheadxi: actually i might be wrong on that... at a glance it doesn't look like gin offers a clean `Close` 😱

postables: :cry:  As mentioned over slack, in another PR we'll do somethign similar to https://github.com/gin-gonic/gin#graceful-restart-or-stop so we can have non-janky+graceful closures.

postables:
This data race travis is reporting is strange as hell. It's not occurring during the PR build, but is only occurring during the branch build. It started with commit https://github.com/RTradeLtd/Temporal/pull/270/commits/9e726f3aa92b6495adde1289d6b90ade09db9e2c which didn't touch anything in the API.

The previous branch build (https://github.com/RTradeLtd/Temporal/pull/270/commits/4c56a29aaad591582e5d3f491bd7cdbba227fb5d) which touched the API, didn't report any issues.

api/v2/api_test.go
```diff
@@ -1832,6 +1833,51 @@ func Test_API_Initialize_Main_Network(t *testing.T) {
...
+	// load configuration
+	cfg, err := config.LoadConfig("../../testenv/config.json")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// setup fake mock clients
+	fakeLens := &mocks.FakeIndexerAPIClient{}
+	fakeOrch := &mocks.FakeServiceClient{}
+	fakeSigner := &mocks.FakeSignerClient{}
```
bobheadxi: this should be in each test case, as declaring outside the scope of a test case means test cases can interfere with each other and thus are no longer isolated

api/v2/api_test.go
bobheadxi: just make sure mocks themselves are instantiated for each test case only (since you set properties on them `ReturnsOnCall` etc and check things like call counts), and are not shared

api/v2/api_test.go
bobheadxi: e.g. https://github.com/RTradeLtd/Lens/blob/master/lens_test.go#L71

bobheadxi: Each test case in this function reuses the same mocks (see line 1861), which we shouldn’t be doing, since it potentially pollutes cases, especially when you test for specific behaviours

bobheadxi: https://github.com/RTradeLtd/Temporal/pull/270/files/236fae1d3ea06ec0ea7965bf5f649c964132372a..f057a05789484ad6f870f163f9576622c029dd66#diff-6e33e0730d3cb4347a70c7d38ce08ef3R1861

api/v2/api_test.go
```diff
@@ -1832,6 +1833,51 @@ func Test_API_Initialize_Main_Network(t *testing.T) {
...
+	type args struct {
+		certFilePath string
+		keyFilePath  string
+	}
+	tests := []struct {
+		name string
+		args args
+	}{
+		{"NoTLS", args{"", ""}},
+		{"TLS", args{"../../testenv/certs/api.cert", "../../testenv/certs/api.key"}},
```
bobheadxi: should add a case for broken files (ie provide the README.md as an argument)

api/v2/routes_rtfsp.go
```diff
@@ -611,12 +611,13 @@ func (api *API) stopIPFSPrivateNetwork(c *gin.Context) {
...
 		return
 	}
-
+	fmt.Println(1)
 	if _, err := api.orch.StopNetwork(c, &ipfs_orchestrator.NetworkRequest{
 		Network: networkName}); err != nil {
 		api.LogError(err, "failed to stop network")(c)
 		return
 	}
+	fmt.Println(2)
```
bobheadxi:
```suggestion

```

api/v2/api_test.go
```diff
@@ -1328,6 +1715,169 @@ func Test_Utils(t *testing.T) {
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			api, err := Initialize(cfg, true, fakeLens, fakeOrch, fakeSigner)
```
bobheadxi:
```suggestion
			fakeLens := &mocks.FakeIndexerAPIClient{}
			fakeOrch := &mocks.FakeServiceClient{}
			fakeSigner := &mocks.FakeSignerClient{}
			api, err := Initialize(cfg, true, fakeLens, fakeOrch, fakeSigner)
```

postables: OH okay I see sorry I misread your comment, good catch

api/v2/api_test.go
postables: I see what you're saying, misread the comment.

api/v2/api_test.go
postables: Good idea

bobheadxi:
@postables looks like Travis is timing out on one of the tests

PR #272, opened 2018-12-15
TEM-154: zappy

postables:

todo

## :rocket: Changes
todo

## :warning: Breaking Changes
todo


api/middleware/jwt.go
```diff
@@ -31,16 +31,10 @@ func JwtConfigGenerate(jwtKey, realmName string, db *gorm.DB, logger *log.Logger
 				return userId, false
 			}
 			if !validLogin {
-				logger.WithFields(log.Fields{
-					"service": "api",
-					"user":    userId,
-				}).Error("bad login")
+				l.With("user", userId).Error("bad login")
```
bobheadxi: this should just be (with the above change):

```suggestion
				l.Error("bad login")
```

api/middleware/jwt.go
```diff
@@ -31,16 +31,10 @@ func JwtConfigGenerate(jwtKey, realmName string, db *gorm.DB, logger *log.Logger
 				return userId, false
 			}
 			if !validLogin {
```
bobheadxi:
```suggestion
			l = l.With("user", userId)
			if !validLogin {
```

api/middleware/jwt.go
```diff
@@ -31,16 +31,10 @@ func JwtConfigGenerate(jwtKey, realmName string, db *gorm.DB, logger *log.Logger
...
-					"user":    userId,
-				}).Error("bad login")
+				l.With("user", userId).Error("bad login")
 				return userId, false
 			}
-			logger.WithFields(log.Fields{
-				"service": "api",
-				"user":    userId,
-			}).Info("successful login")
+			l.With("user", userId).Info("successful login")
```
bobheadxi:
```suggestion
			l.Info("successful login")
```

log/log.go
```diff
@@ -0,0 +1,44 @@
...
+		// Log:         InfoLevel
+		// Encoder:     json
+		// Errors:      stderr
+		// Sampling:    yes
+		// Stacktraces: ErrorLevel
+		config = zap.NewProductionConfig()
+	}
+
+	// set the destination directory for logs
+	config.OutputPaths = append(config.OutputPaths, logPath)
```
bobheadxi: why remove the code for creating the log directory?

api/middleware/middleware_test.go
```diff
@@ -42,8 +41,10 @@ func TestJwtMiddleware(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	logger := log.New()
-	logger.Out = os.Stdout
+	logger, err := log.NewLogger("stdout", true)
```
bobheadxi: `stdout` isn't a real filepath is it?

bobheadxi:
```suggestion
	logger, err := log.NewLogger("", true)
```

api/v2/log.go
```diff
@@ -12,24 +12,25 @@ import (
...
-	if fields != nil {
-		for i := 0; i < len(fields); i += 2 {
-			if i+1 < len(fields) {
-				entry = entry.WithField(fields[i].(string), fields[i+1])
-			}
-		}
-	}
+	logger := api.l
 
 	// write log
```
bobheadxi:
```suggestion
	// write log
	if fields != nil && len(fields)%2 == 0 {
		logger = logger.With(fields...)
	}
```

api/v2/log.go
```diff
@@ -12,24 +12,25 @@ import (
...
 	if err == nil {
-		entry.Error(message)
+		// mod 2 allows us to check the fields are even in length
```
bobheadxi: see above, remove this whole chunk to just be `logger.Error(message`

api/v2/log.go
```diff
@@ -12,24 +12,25 @@ import (
...
+		}
 	} else {
-		entry.WithField("error", err.Error()).Error(message)
+		// mod 2 allows us to check that fields are even in length
+		if fields != nil && len(fields)%2 == 0 {
+			fields = append(fields, "error")
+			fields = append(fields, err.Error())
+			logger.With(fields...).Error(message)
+		} else {
+			logger.With("error", err.Error()).Error(message)
```
bobheadxi: same here, remove this whole chunk and just do a single `logger.Errorw(message, "error", err.Error())`

api/v2/log.go
```diff
@@ -41,22 +42,15 @@ func (api *API) LogError(err error, message string, fields ...interface{}) func(
 
 // LogInfo is a wrapper used by the API to handle simple info logs
 func (api *API) LogInfo(message ...interface{}) {
```
bobheadxi: we can remove these `LogInfo` and `LogDebug` functions entirely now - just use `api.l.Info` and `api.l.Debug`

api/v2/routes_account.go
```diff
@@ -81,23 +80,16 @@ func (api *API) changeAccountPassword(c *gin.Context) {
...
 	} else if !ok {
 		err = fmt.Errorf("password changed failed for user %s to due an unspecified error", username)
 		api.LogError(err, eh.PasswordChangeError)(c)
 		return
 	}
-	api.l.WithFields(log.Fields{
-		"service": api,
-		"user":    username,
-	}).Info("password changed")
+	api.l.With("user", username).Info("password changed")
```
bobheadxi: use `l.Infow` instead

bobheadxi:
```suggestion
	api.l.Infow("password changed",
		"user", username)
```

api/middleware/middleware_test.go
postables: It's not a real file path, I think it's a reserved word by `za`p to log to stdout

bobheadxi: it's already the default

bobheadxi: handled here: https://github.com/RTradeLtd/Temporal/blob/f63e7311fe6f39ff904ae8574c1c53d6d232a169/log/log.go#L37

bobheadxi: see usage in orchestrator: https://github.com/RTradeLtd/ipfs-orchestrator/blob/master/daemon.go#L27 provides a logfile but still goes to stdout

api/v2/log.go
postables: Done!

api/v2/log.go
postables: done

postables:
@bobheadxi you sure? I was going to tackle the queue logging with this PR as well since it's the other place where we extensively use logging. I also want to get rid of that coverage decrease 😅

postables:
Yea that makes sense. I'll merge this then work on the queue logging in the next PR

PR #273, opened 2018-12-16
TEM-154 - 2: Zappy's Back, Tell A Friend

postables:
Replace the queue logger with `zap`
As we don't use shellcheck at the moment, I've temporarily commented the docker pull from the travis file.

## :rocket: Changes
Remove logrus usage from within the queue package, and use `zap` instead.
More significant tests for queues.
The `make testenv` process was slightly updated to download the minio command line client, in order to create the `filesupload` bucket which is used by Temporal.

## :warning: Breaking Changes
None


Makefile
```diff
@@ -52,6 +58,11 @@ testenv:
 	@env CONFIG_DAG=./testenv/config.json go run cmd/temporal/main.go migrate-insecure
 	make api-user
 	make api-admin
+	# create minio bucket
+	wget https://dl.minio.io/client/mc/release/linux-amd64/mc
+	chmod +x mc
+	./mc config host add $(ALIAS) $(HOST) $(ACCESSKEY) $(SECRETKEY)
+	./mc mb $(ALIAS)/$(BUCKET)
```
bobheadxi: can this not be done when Temporal starts up, and use the minio client library we're already using?

postables: I'm not sure where we would do it tbh, this bucket is expected to exist for a lot of tests and from what I can tell from reading docs this cli tool is used for a lot of basic interaction with minio from am administrator perspective.

bobheadxi: It can just be the same as a runmigrations option - if true, init the buckets on startup. Additional binaries are finicky - at the very least the installation here should be removed, and the user should be allowed to decide how to install mc, instead of having the hard coded Linux install here

cmd/temporal/main_test.go
```diff
@@ -69,24 +116,34 @@ func TestQueuesIPFS(t *testing.T) {
...
-		{"IPNSEntry", args{"ipfs", "ipns-entry"}},
-		{"IPFSPin", args{"ipfs", "pin"}},
-		{"IPFSFile", args{"ipfs", "file"}},
-		{"IPFSKey", args{"ipfs", "key-creation"}},
-		{"IPFSCluster", args{"ipfs", "cluster"}},
+		{"IPNSEntry-NoLogDir", args{"ipfs", "ipns-entry", ""}},
+		{"IPNSEntry-LogDir", args{"ipfs", "ipns-entry", "./tmp"}},
+		// omit this test, as ipfs pin spawns a cluster queue publisher
+		// by specifying "", the cluster publisher will attempt to log in the / directory
+		// {"IPFSPin-NoLogDir", args{"ipfs", "pin", ""}},
```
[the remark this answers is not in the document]
postables: Whoops that was due to a slight inconsistency with log dir declaration. I've re-enabled the tests

Makefile
postables: Ah yes good catch, i'll fix this.

cmd/temporal/main.go
```diff
@@ -503,6 +591,8 @@ func main() {
 		}
 		defer signerClient.Close()
 		signer = signerClient
+	case "make-bucket", "make-bucket-insecure":
+		flags["name"] = os.Args[2]
```
bobheadxi: 👌

PR #278, opened 2018-12-21
TEM-172: builds for docker releases

bobheadxi:

See ticket

## :rocket: Changes

- main branch builds (currently `V2` branch) are handled by [Docker Hub](https://cloud.docker.com/u/rtradetech/repository/docker/rtradetech/temporal/hubbuilds)
    - ~version defaults to `latest-{unix-build-time}`~
- ~an additional image will be built in CI on tags that sets version correctly~
- API now contains version info, and reports version on some endpoints
- logger now reports version
- docker hub hooks (see `hooks`)

## :warning: Breaking Changes

stats endpoint now returns stats in `response` instead


bobheadxi:
@postables I'll disable them for now (to be honest... don't think anyone would use those anyway?)

PR #279, opened 2018-12-22
TEM-166: Update IPFS and Other Scripts

postables:
Install scripts were out-dated for multiple packages, also some unused scripts were still around.

## :rocket: Changes
Update IPFS, IPFS Cluster, Grafana, Zabbix, and Prometheus install scripts
Update IPFS manager script to enable namesys pubsub

## :warning: Breaking Changes
None


setup/scripts/prom_node_exporter_install.sh
```diff
@@ -4,7 +4,7 @@
 
 cd ~ || exit
 echo "[INFO] Downloading prometheus onode exporter"
-wget https://github.com/prometheus/node_exporter/releases/download/v0.16.0/node_exporter-0.16.0.linux-amd64.tar.gz
+wget https://github.com/prometheus/node_exporter/releases/download/v0.17.0/node_exporter-0.17.0.linux-amd64.tar.gz
```
bobheadxi: perhaps these versions might be better declared in separate variables just for clarity?

postables: Good idea, I've reflected this across all the scripts, except for the zabbix install one

PR #281, opened 2018-12-25
TEM-141 | API Security Enhancements

postables:
Add security improvements to API where-ever possible
Needs https://github.com/RTradeLtd/database/pull/19 merged
All admin related functionality has been removed from the API, as there are command line tools which satisfy this functionality that are already included in our environments.

## :rocket: Changes
API restriction middleware is removed, as this is a deprecated feature, as authorization can be controlled with `AccountEnabled` field of the user model. This should give some slight performance improvements to API calls, as each API call no-longer needs to perform a database lookup to validate permissions. 

Update CORS to only allow requests from our web-interfaces, as opposed to all origins.

Default dev mode to false.

Replace our security middleware with`gin-contrib/secure`. Complete list of security enhancements:
```Golang
		SSLRedirect:           true,
		IsDevelopment:         false,
		STSSeconds:            315360000,
		STSIncludeSubdomains:  true,
		FrameDeny:             true,
		ContentTypeNosniff:    true,
		BrowserXssFilter:      true,
		ContentSecurityPolicy: "default-src 'self'",
		IENoOpen:              true,
		SSLProxyHeaders:       map[string]string{"X-Forwarded-Proto": "https"},
```

Massive overhaul to the TLS ciphers we accept, as follows:
```Golang
			// configure TLS to override defaults
			tlsConfig := &tls.Config{
				CurvePreferences: []tls.CurveID{tls.CurveP521, tls.CurveP384, tls.CurveP256},
				//PreferServerCipherSuites: true,
				CipherSuites: []uint16{
					// http/2 mandated supported cipher
					// unforunately this is a less secure cipher
					// but specifying it first is the only way to accept
					// http/2 connections without go throwing an error
					tls.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
					// super duper secure ciphers
					tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
					tls.TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA,
					tls.TLS_RSA_WITH_AES_256_GCM_SHA384,
					tls.TLS_RSA_WITH_AES_256_CBC_SHA,
				},
				// consider whether or not to fix to tls1.2
				MinVersion: tls.VersionTLS11,
			}
```
Minio routes have been removed

Any admin functionality has been removed (rtfs cluster, database lookups)

Instead of allowing the user to specify the uploads to search for (this was only done to accomodate admin requests) simply pull the username from the JWT, and use that to define the search scope.

## :warning: Breaking Changes
Basically all non-RTrade owned, and ran web frontends will be broken when using the API, but this isn't of concern, as our own frontends will work. For those running their own temporal instances and a frontend ontop of that, they'll need to update CORS.

The new ciphers will break compatability with a lot of older clients, most notably browser clients, which don't really apply to the API. However anyone unable to use the API with the above accepted ciphers, will be unable to connect. If this becomes a problem, we can adjust the ciphers. However I'm very, very unwilling to do this.

### food for thought

Determine whether we want to disable tls1.1 in favour of minimum tls1.2
Determine whether we want to disable http/2 support in favour of a more secure cipher suite


api/middleware/cors.go
```diff
@@ -5,14 +5,18 @@ import (
...
+
 // CORSMiddleware is used to load our CORS handling logic
-// TODO: need to cleanup and restrict origins, the all origin allow
-// and credentials disable is a temporary measure
 func CORSMiddleware() gin.HandlerFunc {
 	corsConfig := cors.DefaultConfig()
-	corsConfig.AllowAllOrigins = true
-	corsConfig.AllowCredentials = false
+	// configure allowed origins
+	corsConfig.AllowOrigins = allowedOrigins
```
bobheadxi: could allow all for ease of dev work with dev mode enabled

api/v2/api.go
```diff
@@ -257,21 +254,46 @@ func (api *API) Close() {
 	}
 }
 
-// TLSConfig is used to enable TLS on the API service
-type TLSConfig struct {
+// TLSFiles is used to enable TLS on the API service
+type TLSFiles struct {
```
bobheadxi: why change this from `TLSConfig`?

api/v2/routes_account_test.go
```diff
@@ -0,0 +1,220 @@
...
+import (
+	"net/url"
+	"testing"
+
+	"github.com/RTradeLtd/Temporal/mocks"
+	"github.com/RTradeLtd/config"
+	"github.com/RTradeLtd/database/models"
+)
+
+func Test_API_Routes_Account(t *testing.T) {
```
bobheadxi: this test seems long and brittle, it might be a good idea to break it down and move to a more case-driven model

api/v2/api.go
postables: That was a straggler from when I was first experimenting with the TLS cipher configuration. I originally instantiated the http server as:
```Golang
server := &http.Server{
    Addr: ....
    Handler: api.r,
    TLSConfig: ....
```

Which confusing the compiler. Now that the TLS cipher configuration is being setup different then when I first tried, I'll switch it back.

api/middleware/cors.go
postables: good call

api/v2/routes_account_test.go
postables: Good idea.

Gopkg.toml
```diff
@@ -19,15 +19,15 @@ ignored = ["github.com/karalabe/hid", "github.com/satori/go.uuid", "github.com/i
 
 [[override]]
   name = "github.com/RTradeLtd/database"
-  version = "v1.1.1"
```
bobheadxi: going forward, i wonder if we should pin this down to `v1.1.*` (if `dep` even allows version globs like that), since based on our versioning patch increments should maintain full compatibility

api/v2/routes_account_test.go
bobheadxi: this test seems to still be really long and brittle? by case-driven i mean table-driven

it's pretty hard to read, and tests a lot of functionality in one go (**10** endpoints)

Gopkg.toml
postables: I believe with dep we can use constraints like `version = ^1.1.1` or something like that

postables:
For clarity sake, as discussed over slack test case brittleness will be addressed in an upcoming PR adding more API tests.

Gopkg.toml
bobheadxi: this might be a good idea, to avoid having to manually change it every time

PR #289, opened 2019-01-04
api/v2: remove 'api' from route

bobheadxi:

Right now the path looks like:

```
api.temporal.cloud/api/v2
```

the duplicated `api` is kinda awks

## :rocket: Changes

Removes `/api` from the route path

## :warning: Breaking Changes

everything using the API will be broken 😋


postables:
Looks like test is failing on the Lens test, I think the only thing which needs to be fixed is removing the `api` prefix to the call path.

PR #290, opened 2019-01-05
Allow Pinning Of IPNS Hashes

postables:
Previously we assumed all users would give us raw IPFS CIDs to pin. If someone gave us the IPNS hash that was referencing a CID, we would attempt to pin as if it were a CID, not as if it were an IPNS hash.

There is a pretty major overhaul to a package `ipfs/go-ipfs-files` which is required to not use this janky work around, however at the moment that version of IPFS files is incomptabile with the version of IPFS Cluster we are using. The next IPFS Cluster release will use the new `go-ipfs-files` package, so once that is out we'll get rid of this janky method.

## :rocket: Changes
Allow pinning of CIDs references by IPNS records by a dedicated API route which first resolves the IPNS record to extract the hash

## :warning: Breaking Changes
None

### DIFF
--- a/api/v2/routes_rtfs.go
+++ b/api/v2/routes_rtfs.go
@@ -6,13 +6,10 @@ import (
 	"net/http"
 	"strconv"
 
-	path "gx/ipfs/QmZErC2Ay6WuGi96CPg316PwitdwgLo6RxZRqVjJjRj2MR/go-path"
-
 	"github.com/RTradeLtd/Temporal/eh"
 	"github.com/RTradeLtd/Temporal/mini"
 	"github.com/RTradeLtd/Temporal/queue"
 	"github.com/RTradeLtd/Temporal/utils"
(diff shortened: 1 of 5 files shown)


api/v2/routes_rtfs.go
```diff
@@ -61,6 +63,77 @@ func (api *API) pinHashLocally(c *gin.Context) {
...
+		UserName:         username,
+		HoldTimeInMonths: holdTimeInt,
+		CreditCost:       cost,
+	}
+	if err = api.queues.pin.PublishMessageWithExchange(ip, queue.PinExchange); err != nil {
+		api.LogError(err, eh.QueuePublishError)(c)
+		api.refundUserCredits(username, "pin", cost)
+		return
+	}
+	api.l.Infow("ipfs pin request sent to backend", "user", username)
```
bobheadxi: we'll probably want to start attaching logs to some kind of reqeust ID

PR #292, opened 2019-01-07
Add X-Request-ID Header

postables:
Associates a randomly generated `X-Request-Header` to all requests. Anytime an error is logged, the `X-Request-ID` value is logged with the error message to make it easier to troubleshoot issues.

## :rocket: Changes
`X-Request-ID` header inserted with every request, and logged when errors occur

## :warning: Breaking Changes
None

### DIFF
--- a/api/v2/log.go
+++ b/api/v2/log.go
@@ -14,7 +14,7 @@ import (
 // with the given request to make it easier to debug user-submitted erros.
 func (api *API) LogError(c *gin.Context, err error, message string, fields ...interface{}) func(code ...int) {
 	// create base entry with the associated request id
-	logger := api.l.Named(c.GetHeader("X-Request-ID"))
+	logger := api.l.With("request-id", c.GetHeader("X-Request-ID"))
 
 	// write log
 	if fields != nil && len(fields)%2 == 0 {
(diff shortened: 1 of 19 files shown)


api/v2/log.go
```diff
@@ -9,9 +9,12 @@ import (
 // defaults to http.StatusInternalServerError. Fields is an optional set of
 // params provided in pairs, where the first of a pair is the key, and the second
 // is the value
-func (api *API) LogError(err error, message string, fields ...interface{}) func(c *gin.Context, code ...int) {
-	// create base entry
-	logger := api.l
+//
+// Passing in the initial gin.Context to LogError is used to extract the X-Request-ID associated
+// with the given request to make it easier to debug user-submitted erros.
+func (api *API) LogError(c *gin.Context, err error, message string, fields ...interface{}) func(c *gin.Context, code ...int) {
```
bobheadxi: in this case, we don't need the second context anymore

```suggestion
func (api *API) LogError(c *gin.Context, err error, message string, fields ...interface{}) func(code ...int) {
```

api/v2/log.go
```diff
@@ -9,9 +9,12 @@ import (
...
+	// create base entry with the associated request id
+	logger := api.l.Named(c.GetHeader("X-Request-ID"))
```
bobheadxi: `Named` is for components of a system I think. You probably want:
```suggestion
	logger := api.l.With("request-id", c.GetHeader("X-Request-ID"))
```

postables:
This codecov failure is weird, I'm going to merge from the command line

PR #293, opened 2019-01-09
TEM-195: Queue Resiliency

postables:
Occassionally due to connection errors the RabbitMQ consumers and publishers will die, and need to be reconnected. This PR ensures that if the consumers die due to connection errors, they will be recreated.

Allow queue re-building within the API should it also receive connection errors. The one condition with this is that if the re-establish fails, the API process closes. This is the only way I can think of allowing our load-balancer to redirect requests to an API that's not having any issues. From what I can tell, this issue is unlikely to effect both our API instances at once, although that could always happen in which case we should have a service like `monit` on our servers that restarts downed services

## :rocket: Changes
Add function to `Queue::Manager` that creates a channel to receive protocol connection errors on.
In this error is triggered, recreate queue consumers

## :warning: Breaking Changes
None

### DIFF
--- a/api/v2/api.go
+++ b/api/v2/api.go
@@ -318,55 +318,55 @@ func (api *API) ListenAndServe(ctx context.Context, addr string, tlsConfig *TLSC
 		case <-ctx.Done():
 			return server.Close()
 		case msg := <-api.queues.cluster.ErrCh:
-			qmCluster, err := api.handleQueueError(msg, api.cfg, queue.IpfsClusterPinQueue, true)
+			qmCluster, err := api.handleQueueError(msg, api.cfg.RabbitMQ.URL, queue.IpfsClusterPinQueue, true)
 			if err != nil {
 				return server.Close()
 			}
 			api.queues.cluster = qmCluster
 		case msg := <-api.queues.dash.ErrCh:
-			qmDash, err := api.handleQueueError(msg, api.cfg, queue.DashPaymentConfirmationQueue, true)
+			qmDash, err := api.handleQueueError(msg, api.cfg.RabbitMQ.URL, queue.DashPaymentConfirmationQueue, true)
 			if err != nil {
 				return server.Close()
 			}
 			api.queues.dash = qmDash
 		case msg := <-api.queues.database.ErrCh:
-			qmDatabase, err := api.handleQueueError(msg, api.cfg, queue.DatabaseFileAddQueue, true)
+			qmDatabase, err := api.handleQueueError(msg, api.cfg.RabbitMQ.URL, queue.DatabaseFileAddQueue, true)
 			if err != nil {
 				return server.Close()
 			}
 			api.queues.database = qmDatabase
 		case msg := <-api.queues.email.ErrCh:
-			qmEmail, err := api.handleQueueError(msg, api.cfg, queue.EmailSendQueue, true)
+			qmEmail, err := api.handleQueueError(msg, api.cfg.RabbitMQ.URL, queue.EmailSendQueue, true)
 			if err != nil {
 				return server.Close()
 			}
 			api.queues.email = qmEmail
 		case msg := <-api.queues.file.ErrCh:
-			qmFile, err := api.handleQueueError(msg, api.cfg, queue.IpfsFileQueue, true)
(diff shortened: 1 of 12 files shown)


postables:
Here's my thought on how to handle the API side of this:

* When running the API using `ListenAndServe`, we should launch a goroutine for a function named something like `queueRebuilder`. `queueRebuilder` is responsible for watching for closures comming on `amqp.Error` channel for each seperate queue process. If an error is detected, we then rebuild the failed queue.

api/v2/api.go
```diff
@@ -315,6 +315,132 @@ func (api *API) ListenAndServe(ctx context.Context, addr string, tlsConfig *TLSC
...
+				"error", msg.Error())
+			qmPay, err := queue.New(queue.PaymentConfirmationQueue, api.cfg.RabbitMQ.URL, true, api.l)
+			if err != nil {
+				api.l.Errorw(
+					"failed to re-establish queue process, exiting",
+					"queue", "pay-confirm",
+					"error", err.Error())
+				return server.Close()
+			}
+			api.queues.payConfirm = qmPay
```
bobheadxi: a lot of repetition here - maybe something like:

```go
// this doesnt exist yet but would be nice to have
type Queue string

// this doesnt exist yet but would be nice to have
type queue.Options{
   mqUrl string
   publish bool
   logger *zap.Logger
}

func handleQueueError(err error, m *queue.Manager, q queue.Queue, opts queue.Options) error {
	api.l.Errorw("a protocol connection error stopping rabbitmq was received",
				"queue", q,
				"error", msg.Error())
         manager, err = queue.New(q, opts) // assigns to appropriate queue
        return err
}
```

queue/types.go
```diff
@@ -64,6 +67,7 @@ type Manager struct {
 	l            *zap.SugaredLogger
 	db           *gorm.DB
 	cfg          *config.TemporalConfig
+	ErrChannel   chan *amqp.Error
```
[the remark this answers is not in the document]
postables: Good suggestion

cmd/temporal/main.go
```diff
@@ -298,23 +338,31 @@ var commands = map[string]cmd.Cmd{
...
+						}
+						waitGroup.Add(1)
+						if err := qm.ConsumeMessages(ctx, waitGroup, db, &cfg); err != nil && err.Error() != queue.ReconnectError {
+							fmt.Println("failed to consume messages", err)
+							os.Exit(1)
+						} else if err != nil && err.Error() == queue.ReconnectError {
+							continue
+						}
+						// this will only be true if we had a graceful exit to the queue process, aka CTRL+C
+						if err == nil {
```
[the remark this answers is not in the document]
postables: WHOOPS.... I'll have to change that in multiple places in `cmd/temporal/main.go` since i copy pasted this statement.

api/v2/api.go
postables: very nice will implement :ok_hand:

postables: Ok I implemented like 75% of this suggestion. I think we might be able to do something with our `API::Queues` struct, and have each individual queue be a `QueueSettings` or something but i'll link into this and if it is possible, do it in a separate PR.

api/v2/api.go
```diff
@@ -566,3 +622,22 @@ func (api *API) setupRoutes() error {
 	api.l.Info("Routes initialized")
 	return nil
 }
+
+// HandleQueueError is used to handle queue connectivity issues requiring us to re-connect
+func (api *API) handleQueueError(amqpErr *amqp.Error, cfg *config.TemporalConfig, queueType queue.Queue, publish bool) (*queue.Manager, error) {
```
bobheadxi: I think `config.TemporalConfig` is wayyyy to broad of a scope to be passing around here - if anything, replace it with just `string` for now

where possible, we want to narrow the scope of what gets passed around so that each function takes as little as it needs to be functional, that way its dependencies are clear. without digging into this code, it's impossible to know whether `handleQueueError` relies on every single field in config, half of them, or (in this case) just one

postables: good point!

postables:
Thanks! I'm glad to finally solve this issue it's cropped up like once a month. Last time was November, good timing 😂

PR #297, opened 2019-01-19
Adds Better Monitoring And Alerting

postables:
Update IPFS and IPFS Cluster monitoring scripts
Updates IPFS Zabbix template to include all monitoring functionality, triggers to generate alerts, graphs, and screen for condensed graph viewing

Triggers will be fired, and send emails when:
* Queued pins on ipfs cluster go above a configurable threshold
* IPFS datastore is detected as not writable
* IPFS Cluster daemon status experiences an issue
* IPFS peer count drops below a configurable threshold
* IPFS Cluster pin errors goes above a certain threshold
* IPFS Daemon goes offline

TODO:
* Do the same but for temporal services
* Add `monit` configuration to restart failed IPFS daemons followed by an IPFS Cluster daemon restart

## Notes

* Designed to work with Zabbix Agent passive checks, eventually will move to active checks.
* The zabbix user needs to be able to write to `$IPFS_DATASTORE_PATH`. Traditional `chmod` is not-desirable because it grants pretty loose permissions, the better option is to run the following command to only grant write permission `sudo setfacl -m u:zabbix:w "$IPFS_DATASTORE_PATH"`. 
* Possible alternative: create a directory within the `$IPFS_DATASTORE_PATH` used to test, that way the zabbix agent can't write to the main directory. Not exactly the biggest issue since we aren't granting read permissions
* The `ipfs_monitor.sh` script contains declarations for `$IPFS_PATH` and `$IPFS_DATASTORE_PATH` as I don't want to grant those to `zabbix` user out-side of the context of the monitor lookup.


postables:
pretty sure I found the issue causing a data race during the test, `cmd/temporal/main_test.go` line 53 we launch multiple APIs in go routines. As the  gRPC clients are initially declared as package level variables, the tests will race (I think). although a little strange this isn't occuring in the PR build.

on the bus home, will try that fix when I'm home.

PR #303, opened 2019-02-06
Key Exchange Fix: TEM-223

postables:
Key creation messages are supposed to be sent to all active consumers, instead they were only being sent to one.

## :rocket: Changes
The problem was that all key consumers were registering under the same queue, which with the fanout exchange effectively because round-robin distribution, as there is a single queue being registered under exchange. The fix was to generate a some-what unique queue name, and bind that to the exchange.  This then allow the fanout exchange to route to all the queues attached to the exchange.

## :warning: Breaking Changes
None

### DIFF
--- a/queue/queue.go
+++ b/queue/queue.go
@@ -7,14 +7,29 @@ import (
 	"encoding/json"
 	"errors"
 	"io/ioutil"
+	"os"
 	"sync"
 
+	"github.com/RTradeLtd/Temporal/utils"
 	"github.com/RTradeLtd/gorm"
 	"go.uber.org/zap"
 
 	"github.com/RTradeLtd/config"
 	"github.com/streadway/amqp"
 )
+
+func (qm *Manager) parseQueueName(queue Queue) error {
+	host, err := os.Hostname()
+	if err != nil {
+		return err
+	}
+	// the following is to account for running multiple queue workers
+	// on a single host.
+	rand := utils.GenerateRandomUtils()
+	s := rand.GenerateString(5, utils.LetterBytes)
+	qm.QueueName = queue + Queue(host+s)
+	return nil
+}
 
 // New is used to instantiate a new connection to rabbitmq as a publisher or consumer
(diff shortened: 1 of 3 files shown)


queue/queue.go
```diff
@@ -34,21 +49,25 @@ func New(queue Queue, url string, publish bool, cfg *config.TemporalConfig, logg
...
+		if err := qm.parseQueueName(queue); err != nil {
+			return nil, err
+		}
+		// setup the actual exchange
 		if err = qm.setupExchange(queue); err != nil {
 			return nil, err
 		}
 	}
+	// if we aren't publishing, and are consuming
+	// setup a queue to receive messages on
```
```suggestion
	// set up a queue to receive messages on
```

queue/queue.go
```diff
@@ -34,21 +49,25 @@ func New(queue Queue, url string, publish bool, cfg *config.TemporalConfig, logg
...
-		}
-	}
 	// Declare Non Default exchanges for matching queues
 	switch queue {
 	case IpfsKeyCreationQueue:
+		// as this is an exchange based queue we need a unique queue name
+		if err := qm.parseQueueName(queue); err != nil {
+			return nil, err
+		}
+		// setup the actual exchange
```
```suggestion
		// set up the actual exchange
```

queue/queue_test.go
```diff
@@ -22,6 +24,42 @@ const (
...
+		t.Fatal("failed to properly parse queue name")
+	}
+}
+
+func TestParseQueueFull(t *testing.T) {
+	cfg, err := config.LoadConfig(testCfgPath)
+	if err != nil {
+		t.Fatal(err)
+	}
+	loggerConsumer, err := log.NewLogger("", true)
```
bobheadxi: I recently learned about this: https://godoc.org/go.uber.org/zap/zaptest#NewLogger

It logs all logs to `t.Log` instead, so only logs in the specific failing test shows up, not urgent now but at least for the API tests it will help with debugging

postables: interesting, that will be very helpful

PR #307, opened 2019-02-14
Fix Mismatching IPFS Key Generation

postables:
Currently a key creation request is serviced by two queue runners simultaneously. However due to the random nature of key generation,  this would result in the same key name having two different content hashes, but only one of them stored in the database. On subsequent IPNS record creation. the hash of the record would be the hash of the key on whatever queue runner serviced the request, but there is a 50% chance the key used, wouldn't match what was in the database. Thus, although the IPNS records would be published it might not match what is in the database.

## :rocket: Changes
* remove the ipfs key creation exchange
* instantiate two versions of our kaas client in our key creation queue
* distribute key creation requests round robin to either of our queue runners
* the processing queue runner generates the key, and stores it in both instances of kaas.

## :warning: Breaking Changes
none

## post-merge tasks:

release a new version


CHANGELOG-V2.md
```diff
@@ -3,6 +3,42 @@
 This document tracks changes to Temporal and its related projects for all `v2.x.x`
 releases. See our [versioning policy](/VERSIONING.md) for more details.
 
+## v2.0.4
+
+* relevant PRs:
+  * [#305](https://github.com/RTradeLtd/Temporal/pull/305)
+  * [#306](https://github.com/RTradeLtd/Temporal/pull/306)
+  * [#307](https://github.com/RTradeLtd/Temporal/pull/307)
+  * [#308](https://github.com/RTradeLtd/Temporal/pull/308)
```
bobheadxi: 🤔 this is fine i guess but in the future we should try to make sure each fix gets its own PR, so that people can better see changes for themselves and vet them if they want

Makefile
```diff
@@ -107,6 +107,26 @@ vendor:
 	# Nuke vendor directory
 	rm -rf vendor
 
+	# rebuild dependencies
+	dep ensure -v
```
bobheadxi: instead of a separate `vendor-update` script, maybe we could just do:

```suggestion
	dep ensure -v $(DEPFLAGS)
```

then:

```
make vendor DEPFLAGS=-update
```

CHANGELOG-V2.md
postables: this is a good idea, will do this from now on :ok_hand:

Makefile
postables: neat!

grpc-clients/kaas.go
```diff
@@ -0,0 +1,67 @@
...
+				grpc.WithTransportCredentials(creds),
+				grpc.WithPerRPCCredentials(dialer.NewCredentials(opts.Krab.AuthKey, true)))
+		} else {
+			dialOpts = append(dialOpts,
+				grpc.WithInsecure(),
+				grpc.WithPerRPCCredentials(dialer.NewCredentials(opts.Krab.AuthKey, false)))
+		}
+		url = opts.Krab.Fallback.URL
+	} else {
+		if opts.Krab.TLS.CertPath != "" {
```
bobheadxi: why duplicate all this code? it seems to me the only difference is the TLS config - maybe we could just assign the appropriate config based on `fallback`

postables: good point, how does this look
```golang
// NewKaasClient is used to instantiate our kaas client in primary or fallback mode
func NewKaasClient(opts config.Services, fallback bool) (*KaasClient, error) {
	var (
		dialOpts   []grpc.DialOption
		krabConfig config.Krab
	)
	if fallback {
		krabConfig.AuthKey = opts.Krab.Fallback.AuthKey
		krabConfig.KeystorePassword = opts.Krab.Fallback.KeystorePassword
		krabConfig.LogFile = opts.Krab.Fallback.LogFile
		krabConfig.TLS = opts.Krab.Fallback.TLS
		krabConfig.URL = opts.Krab.Fallback.URL
	} else {
		krabConfig = opts.Krab
	}
	if krabConfig.TLS.CertPath != "" {
		creds, err := credentials.NewClientTLSFromFile(krabConfig.TLS.CertPath, "")
		if err != nil {
			return nil, fmt.Errorf("could not load tls cert: %s", err)
		}
		dialOpts = append(dialOpts,
			grpc.WithTransportCredentials(creds),
			grpc.WithPerRPCCredentials(dialer.NewCredentials(krabConfig.AuthKey, true)))
	} else {
		dialOpts = append(dialOpts,
			grpc.WithInsecure(),
			grpc.WithPerRPCCredentials(dialer.NewCredentials(krabConfig.AuthKey, false)))
	}
	conn, err := grpc.Dial(krabConfig.URL, dialOpts...)
	if err != nil {
		return nil, err
	}
	return &KaasClient{
		conn:          conn,
		ServiceClient: pb.NewServiceClient(conn),
	}, nil
}
```

bobheadxi: I notice [here](https://github.com/RTradeLtd/config/blob/master/types.go#L175) that fallback is *exactly* the normal config - why not just make `Fallback` of type `Krab` and here we can have:

```go
if fallback {
    conf = opts.Krab.Fallback
} else {
    conf = opts.Krab
}
```

postables: even better!

postables: hmm that would lead to a recursive type

bobheadxi: probably dont put it together then

```
opts.Krab
opts.KrabFallback
```

postables: Excellent idea, heading to bed right now but will do this in the morning 👌

postables: Ok made the change, also realized one other spot where inconsistencies would show up during key export, so I configure the API to have clients for both kb1+kb2 and issue deletes to both instances

PR #308, opened 2019-02-17
Account Restrictions

postables:
Adds a couple of account restrictions, and lays the foundation for deduplicated storage billing, as well as a few other helpers

## :rocket: Changes
* limit free accounts to 1 month pin times
* limit non-free accounts to 2 year pin times
* add api call to extend pin time
* prevent accounts for re-uploading the same data multiple times to avoid duplicate billing, instead fail gracefull with a message suggesting they extend pin time
* adds the foundation for per-customer deduplicated storage billing

## :warning: Breaking Changes
None

## post merge tasks:
update api docs
hold off on new release until i can add links to proper api terms and services (eta 2-3 days) this will be introduced in a separate pr


api/v2/routes_account.go
```diff
@@ -143,9 +142,24 @@ func (api *API) registerUserAccount(c *gin.Context) {
 	// log
 	api.l.With("user", forms["username"]).Info("user account registered")
 	// remove hashed password from output
-	userModel.HashedPassword = "scrubbed"
+	user.HashedPassword = "scrubbed"
+	// remove the verification token from output
+	user.EmailVerificationToken = "scrubbed"
+	// format a custom response that includes the user model
+	// and an additional status field
+	type userRegisteredModel struct {
```
bobheadxi: why not just declare this inline at `gin.H`?

api/v2/routes_ipns.go
```diff
@@ -153,6 +153,20 @@ func (api *API) pinIPNSHash(c *gin.Context) {
 	// IPFS Cluster doesn't, so to keep consistency with the rest of our endpoints
 	// we should only operate on a bare cidC
 	hash := strings.Split(hashToPin, "/")[len(strings.Split(hashToPin, "/"))-1]
+	upload, err := api.upm.FindUploadByHashAndUserAndNetwork(username, hash, "public")
+	// by this conditional if statement passing, it means the user has
+	// upload content matching this hash before, and we don't want to charge them
+	// so we should gracefully abort further processing
+	if err == nil || upload != nil {
+		Respond(
+			c,
```
bobheadxi: this is really

api/v2/routes_ipns.go
```diff
@@ -153,6 +153,20 @@ func (api *API) pinIPNSHash(c *gin.Context) {
...
+			http.StatusOK,
+			gin.H{
```
bobheadxi: this is really spread out 😋 I would try to compress it a bit for readability:

```go
Response(c,
   http.StatusOK,
   gin.H{
      // ...
   })
```

api/v2/routes_rtfs.go
```diff
@@ -311,5 +354,73 @@ func (api *API) getDagObject(c *gin.Context) {
...
+	// hash to retrieve dag for
+	hash := c.Param("hash")
+	if _, err := gocid.Decode(hash); err != nil {
+		Fail(c, err)
+		return
+	}
+	// extract needed post forms
+	forms := api.extractPostForms(c, "hold_time")
+	if len(forms) == 0 {
+		return
```
bobheadxi: don't we need a proper call to `Fail` or something of the sort for the request to end properly?

customer/types.go
```diff
@@ -0,0 +1,26 @@
...
+
+	"github.com/RTradeLtd/database/models"
+	"github.com/RTradeLtd/rtfs"
+)
+
+// Manager is used to handle managing customer objects
+type Manager struct {
+	um    *models.UserManager
+	ipfs  rtfs.Manager
+	mutex sync.Mutex
```
bobheadxi: is this used yet?

api/v2/routes_rtfs.go
postables: Hmm that's a good point, I suppose this is a bit unclear actually since the "failure" is sent within `api.extractPostForms`

```golang
func (api *API) extractPostForms(c *gin.Context, formNames ...string) map[string]string {
	forms := make(map[string]string)
	for _, name := range formNames {
		value, exists := c.GetPostForm(name)
		if !exists {
			FailWithMissingField(c, name)
			return nil
		}
		forms[name] = value
	}
	return forms
}
```

let me refactor this a bit to be more clear

api/v2/routes_account.go
postables: good point :man_facepalming:

postables: Actually on second thought, I'm not sure this would work? :thinking: `gin.H` expects a `string -> interface`, so the only way I can think of doing this would be:
```golang
userRegisteredModel := gin.H{"user": user, "status": status}
```

This would then break compatibility with the current response expected from this api call?

customer/types.go
postables: the mutex or the customer object manager? Originally I was intending for customer object manager to be released with this PR, but it ended up being a bit more work then I expected so I want to target releasing this functionality in the next release after this one. So technically the customer object manager isn't being used yet. Mutex is being used when running `Manager::Update`

api/v2/routes_ipns.go
postables: yes that is a good point.

api/v2/routes_rtfs.go
bobheadxi: yeah functions with hidden side effects like that are probably a bad idea - i would return an error with `extractPostForms` isntead

bobheadxi: oh your new implementation with `missingField` seems okay too 🆗

customer/types.go
bobheadxi: We should not be putting core functions of a class in separate files from the definition

bobheadxi: ie as much as possible, `Manager`'s public functions should be in this file where `Manager` is defined

postables: that is a good point :ok_hand:

api/v2/routes_account.go
bobheadxi:
```go
gin.H{"response": gin.H{ "status": status, "user": user } })
```

bobheadxi: oh i kinda see what you mean - you're adding a new field to user

you can declare an [anonymous struct](https://talks.golang.org/2012/10things.slide#2) inline in the response then

bobheadxi: like so: https://play.golang.org/p/60n6nR4iBo6

postables: shit, thats awesome

postables:
note: of the 4 PRs that were opened, i'll be merging this last, with an update to the changelog

PR #311, opened 2019-02-20
Allow Uploading Directories

postables:
Allow uploading directories, particularly useful for adding websites to ipfs

## :rocket: Changes
* Adds API call that takes in a zip file, unzips, and uploads the directory to ipfs

## :warning: Breaking Changes
none


postables:
The test is being weird, and when running in travis it reports that the zip file isn't actually a zip file?? not sure how to fix

api/v2/utils.go
```diff
@@ -205,3 +210,65 @@ func (api *API) extractPostForms(c *gin.Context, formNames ...string) map[string
 	}
 	return forms
 }
+
+// Unzip will decompress a zip archive, moving all files and folders
+// within the zip file (parameter 1) to an output directory (parameter 2).
+// from https://golangcode.com/unzip-files-in-go/
+func Unzip(src string, dest string) ([]string, error) {
```
bobheadxi: is this safe, to unzip arbitrary zip files - or files that the uploader claim is a zip file - onto our servers?

postables: There are a few security issues, I'm going to add a few more checks and balances here. The main issue I knew of was zip slip which this protects against, however there are some additional issues I didnt think of.

postables: One of the things we might need to change is running the API docker using [gvisor](https://github.com/google/gvisor)

postables:
@bobheadxi I have added a bit extra protections to the API call to protect against a zip bomb attack, however to protect against malware and other malicious attacks, we'll need to run the API in a sandboxed environment. We should be able to accomplish this using [gvisor](https://github.com/google/gvisor) with our API docker conatainer.

api/v2/routes_rtfs.go
```diff
@@ -279,6 +280,29 @@ func (api *API) uploadDirectory(c *gin.Context) {
...
+		Fail(c, err)
+		return
+	}
+	// protect against zip bombs
+	var uncompressedSize uint64
+	for _, f := range z.File {
+		uncompressedSize = uncompressedSize + f.UncompressedSize64
+		if uncompressedSize >= datasize.GB.Bytes() {
+			z.Close()
+			Fail(c, errors.New("uncompressed size of zip file is larger than 1gb max upload"))
```
bobheadxi: 🤖 nice nice

api/v2/utils.go
```diff
@@ -205,3 +210,65 @@ func (api *API) extractPostForms(c *gin.Context, formNames ...string) map[string
...
+			if err = os.MkdirAll(filepath.Dir(fpath), os.ModePerm); err != nil {
+				return filenames, err
+			}
+
+			outFile, err := os.OpenFile(fpath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, f.Mode())
+			if err != nil {
+				return filenames, err
+			}
+
+			_, err = io.Copy(outFile, rc)
```
bobheadxi: might want to check the number of bytes written, so if its `0` something probably went bad

postables: good idea

PR #312, opened 2019-02-21
Enable gVisor Runtime (WIP)

postables:
The introduction of [pull/311](https://github.com/RTradeLtd/Temporal/pull/311) introduces a potential security risk, this PR is to address that, my enabling the sandbox of the API docker container to use the gvisor runtime.

## :rocket: Changes
* Update makefile to download, and configure basic gvisor runtime for docker
* Add target to makefile to set docker daemon to use runsc as the default runtime

## :warning: Breaking Changes
None


postables:
@bobheadxi any idea how we can go about enbaling the `runsc` runtime for `temporal.yml`?

postables:
Looks like we can set `runsc` as the default runtime:
https://stackoverflow.com/questions/47465696/how-do-i-specify-nvidia-runtime-from-docker-compose-yml

.gitignore
```diff
@@ -7,3 +7,5 @@
 tmp
 
 \.vscode/
+
+runsc\.sha512
```
bobheadxi: why not `wget` this file into `tmp` so we don't need to have it lying around in root?

setup/configs/docker/README.md
```diff
@@ -0,0 +1,6 @@
+# docker
+
+Configuration files for docker
+
+`daemon.json` is docker daemon setup for basic gvisor runtime usage
+`daemon_passthrough.json` is the same as `daemon.json` but configured to allow network passthrough as the risk of reduced isolation
```
bobheadxi:
```suggestion
`daemon_passthrough.json` is the same as `daemon.json` but configured to allow network passthrough at the risk of reduced isolation
```

perhaps just specify that this is done through using host network mode, "network passthrough" seemed a little confusing (ie rename this `daemon_hostnetwork.json`)

postables:
yes, I think even if this unzip api call for directory uploads isn't used (after i've been thinking about it longer I'm a little skeptical of using it), gvisor is a pretty good thing to implement since Temporal handles a lot of user uploaded data.

Makefile
```diff
@@ -142,6 +147,17 @@ release-docker: docker
 	@docker push rtradetech/temporal:$(TEMPORALVERSION)
 	@echo "===================          done           ==================="
 
+# download and setup gvisor runtime
+.PHONY: gvisor
+gvisor:
+	wget https://storage.googleapis.com/gvisor/releases/nightly/latest/runsc -O /tmp/runsc
```
bobheadxi: `./tmp` is what i was thinking 😋general rule of thumb, try not to fiddle with root-level directories

bobheadxi: (since `./tmp` is in `gitignore`)

postables: hah good point :man_facepalming:

PR #313, opened 2019-02-22
Fix Krab In Development And Add Proper Resiliency

postables:
Krab wasn't working in development environment due to opening up two grpc connections to same server from same host, also there wasn't any true resiliency as a failure to store/get krab from a single instnace would abort processing

## :rocket: Changes
* only instantiate backup krab in non dev mode
* only invoke backup krab commands in non dev mode
* when publishing an IPNS record and failing to retrieve key from primary krab, try backup, if backup fails abort

## :warning: Breaking Changes
None


queue/queue.go
```diff
@@ -111,7 +111,8 @@ func (qm *Manager) declareQueue() error {
 // ConsumeMessages is used to consume messages that are sent to the queue
 // Question, do we really want to ack messages that fail to be processed?
 // Perhaps the error was temporary, and we allow it to be retried?
-func (qm *Manager) ConsumeMessages(ctx context.Context, wg *sync.WaitGroup, db *gorm.DB, cfg *config.TemporalConfig) error {
+func (qm *Manager) ConsumeMessages(ctx context.Context, wg *sync.WaitGroup, db *gorm.DB, cfg *config.TemporalConfig, dev bool) error {
```
bobheadxi: why is this per-`ConsumeMessages`? why cant it be set as part of `Manager`?

postables: good point what would be simpler

PR #315, opened 2019-02-25
Add Stripe Payments

postables:
* Allow purchasing credits with stripe
* ~First set of refactor to API tests~ postpone test refactors to additional pr

## :rocket: Changes
* Adds a stripe payment api call
* stripe-go dependency added

## :warning: Breaking Changes
* None


api/v2/api.go
```diff
@@ -433,6 +433,10 @@ func (api *API) setupRoutes() error {
 			eth.POST("/request", api.RequestSignedPaymentMessage)
 			eth.POST("/confirm", api.ConfirmETHPayment)
 		}
+		stripe := payments.Group("/stripe")
+		{
+			stripe.POST("/charge", api.stripeCharge)
```
bobheadxi: is there a reason where the payments are:

* dash -> create
* eth -> request
* stripe -> charge

why not:

```
/payments/charge/${method}
```

postables: Yes:

ethereum for example is a two step process:
1) you need to `request` a signed message
2) submit the signed message to a smart contract and `confirm` it

dash involves `create`ing a payment forward with https://chainrider.io/ in the future dash payment methods will allow for reusable payment forward ids, which may end up having a call like `dash/reuse`.

Additionally I want to have the payment methods grouped by type, as each method might have different management/usability functions which aren't necessary related to different ones. For example with stripe you are placing a `charge` on a credit card.

bobheadxi: I see - it seemed like a missed opportunity to build an abstraction, but I suppose if we're aiming to keep it low level then this is alright

postables: Depending on how the payment system evolves it could definitely be revamped in the future when an eventual breaking change has to happen.

postables:
@bobheadxi I'm looking to *finally* get the API test refactor done. Can you take a look at the tests in `routes_rtfs_test.go` and let me know how they look? If they've got your approval, I'm going to continue with the refactor.

postables:
test refactoring was more involved than i expected it to be, and is being done in a separate branch https://github.com/RTradeLtd/Temporal/tree/tests/refactor

postables:
broke this branch pretty badly due to some overly eager test refactoring, opening up a new pr

PR #325, opened 2019-03-20
Fix Timeouts When Virus Scanning Directory Uploads

postables:
When uploading a zipped file of a significant size, clamd would timeout and fail.

## :rocket: Changes
* Adjust connection and cmd timeouts for the clamd utility in Temporal
* Make sure to adjust settings on clamd configuration server side

## :warning: Breaking Changes
None

### DIFF
--- a/api/v2/routes_rtfs.go
+++ b/api/v2/routes_rtfs.go
@@ -354,7 +354,7 @@ func (api *API) uploadDirectory(c *gin.Context) {
 	if err := z.Close(); err != nil {
 		// remove file if this fails
 		os.Remove(destPathZip)
-		api.LogError(c, err, err.Error())(http.StatusInternalServerError)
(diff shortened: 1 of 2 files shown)


api/v2/routes_rtfs.go
```diff
@@ -354,7 +354,7 @@ func (api *API) uploadDirectory(c *gin.Context) {
 	if err := z.Close(); err != nil {
 		// remove file if this fails
 		os.Remove(destPathZip)
-		Fail(c, err)
+		api.LogError(c, err, err.Error())(http.StatusInternalServerError)
```
bobheadxi: it might be safe to *not* return an error here - it's not relevant to the user whether the file was removed serverside, and we'll see the error in our logs

postables: I returned a generic error message, I don't think we can avoid aborting processing here, since it would mean that we aren't closing the read handler we have on the zip file.

PR #326, opened 2019-04-06
TEM-157: scaffold v3 structure, implement authentication

bobheadxi:

This is the first step of the Temporal API v3 project, and lays out:

* RPC definition (at https://github.com/RTradeLtd/sdk)
* structure for service implementation (see `AuthService` and stubs for the other services)
* structure for dependency interfaces (see `interfaces.go`)
* middleware, context values, etc.

## :rocket: Changes

* set up scripts, directories, packages, etc. for the v3 API
* implement `AuthService`

## :warning: Breaking Changes

none, hopefully


bobheadxi:
@postables seems like there is an unrelated test failing - https://travis-ci.com/RTradeLtd/Temporal/builds/107309679#L624, any idea why?

postables:
@bobheadxi that test is failing due to an out-of-date IPFS Cluster version. Looks like the test environment is using [`v0.8.0`](https://github.com/RTradeLtd/testenv/blob/174a3ea4e855fb41c844f95b68843c9259f28484/docker-compose.yml#L21) when it needs to be at `0.10.0`

api/v3/auth.go
```diff
@@ -0,0 +1,612 @@
...
+		}); err != nil {
+			l.Errorw("failed to send email to queue",
+				"error", err)
+			return nil, grpc.Errorf(codes.Internal, eh.QueuePublishError)
+		}
+		return &auth.Empty{}, nil
+
+	case auth.RecoverReq_USERNAME:
+		if err := a.emails.PublishMessage(queue.EmailSend{
+			Subject:     "TEMPORAL User Name Reminder",
```
postables: I think we should probably rename these instances of `TEMPORAL` to `Temporal`

api/v3/auth.go
```diff
@@ -0,0 +1,612 @@
...
+func (a *AuthService) Register(ctx context.Context, req *auth.RegisterReq) (*auth.User, error) {
+	var (
+		email = req.GetEmailAddress()
+		user  = req.GetCredentials().GetUsername()
+		pw    = req.GetCredentials().GetPassword()
+		l     = a.l.With("user", user, "email", email)
+	)
+
+	// validate email
+	if strings.ContainsRune(email, '+') {
```
postables: This reminds me as well, we should probably also validate the email is valid as well. The inability to use a `+` was due to email catch-all routing, but I've noticed people signing up with emails like `username.............` so if we get repeated sequences of a `.` its probably safe to deem the email as invalid as someone either typo'd or is just being a troll

Basically the validation rules are:
* No usage of catch-all routing (+)
* No trailing periods (ie, `username.@example.org`, `username..........@example.org`)

api/v3/auth.go
```diff
@@ -0,0 +1,612 @@
...
+
+	// streamInterceptor handles all incoming stream RPC requests
+	streamInterceptor = func(
+		srv interface{},
+		stream grpc.ServerStream,
+		info *grpc.StreamServerInfo,
+		handler grpc.StreamHandler,
+	) (err error) {
+		if v, found := exclude[info.FullMethod]; !v && !found {
+			a.l.Debugw("requested RPC is an exception - skipping authentication",
```
postables: What is this doing, and to what does it apply to? looks like a good place to add a quick explainer comment

api/v3/auth.go
bobheadxi: Did the above, and also implement a basic domain verification - see 5d52f34

api/v3/auth.go
bobheadxi: 8b80cc4 - please let me know if this doens't clarify it! The tests also demonstrate what is going on:

https://sourcegraph.com/github.com/RTradeLtd/Temporal@api/v3/prototype/-/blob/api/v3/auth_test.go#L913:57

api/v3/auth.go
bobheadxi: a2be798

bobheadxi:
> @bobheadxi that test is failing due to an out-of-date IPFS Cluster version. Looks like the test environment is using v0.8.0 when it needs to be at 0.10.0

seems strange, when did we upgrade? https://github.com/RTradeLtd/testenv/blob/master/docker-compose.yml#L21 was never updated

postables:
@bobheadxi whoops that was my mistake, I updated testenv but forgot to merge https://github.com/RTradeLtd/testenv/tree/ipfs-cluster

api/v3/auth.go
postables: Looks great!

api/v3/auth.go
postables: Excellent :ok_hand:

api/v3/auth.go
postables: :fire:

bobheadxi:
update on this:

* [ ] gateway needs some work and refactoring
* [ ] need to make sure everything can get up and running

PR #328, opened 2019-04-09
User Management For Private Networks And Access Restrictions

postables:
If a user was authorized to access a private network they could invoke destructive commands such as destroying the network. Additionally, you were unable to modify authorized users after the network is created.

Needs [database/33](https://github.com/RTradeLtd/database/pull/33) to be merged first

## :rocket: Changes
* Add API call to add+remove user(s) to/from network access list
* Restrict all network management commands to be access by the owner only
* Allow for setting multiple owners

### DIFF
--- a/api/v2/routes_rtfsp_management.go
+++ b/api/v2/routes_rtfsp_management.go
@@ -103,31 +103,14 @@ func (api *API) startIPFSPrivateNetwork(c *gin.Context) {
 	}
 	logger := api.l.With("user", username, "network_name", networkName)
 	logger.Info("private ipfs network start requested")
-	// get all networks user has access too
-	networks, err := api.um.GetPrivateIPFSNetworksForUser(username)
+	// verify admin access to network
+	network, err := api.nm.GetNetworkByName(networkName)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


api/v2/routes_rtfsp_management.go
```diff
@@ -354,3 +301,174 @@ func (api *API) getAuthorizedPrivateNetworks(c *gin.Context) {
...
+			api.LogError(c, err, "failed to update network for user")(http.StatusInternalServerError)
+			return
+		}
+	}
+	Respond(c, http.StatusOK, gin.H{"response": "authorized user list updated"})
+}
+
+// removeUsersFromNetwork is used to remove a user from being able
+// to access a network.
+func (api *API) removeUsersFromNetwork(c *gin.Context) {
```
bobheadxi: `api` exposes a huge number of functions 😱 

something I'm hoping to change with V3! but for now all good 🚀

postables: Agreed haha, that's one of the things I'm looking forward to about the V3 API you're working on :D

PR #333, opened 2019-05-15
Enable Bitcoin Cash Payments (WIP)

postables:

Enables BCH as a supported crypto payment method. This will also lay 75% of the foundation needed to accept BTC as a supported crypto payment method, so it is essentially killing 2 birds with one stone.

## :rocket: Changes

* Enable coinmarketcap price lookups of BCH
* Remove function `validateBlockchain` that was only being used in a test within the v2 api.

## :warning: Breaking Changes

None


postables:
In order to successfully handle payments, each payment creation will need to use a unique address. This will involve leveraging `bchwallet` to create handle management of a bchwallet.

postables:
Strange, the test build I did for RC1 didnt seem to pickup the API calls for BCH

postables:
Current method of generating an address for payment can be better optimized. See [this](https://github.com/gcash/bchwallet/issues/41) issue for more information

Issue #365, opened 2019-09-16
2 critical issues requesting POST DOWNLOAD API endpoint with native fetch or xhr javascript browser


chisNaN:
2 critical issues requesting POST DOWNLOAD API endpoint with native fetch or xhr javascript browser

**1) First problem concerns this endpoint url https://api.temporal.cloud/v2/ipfs/utils/download/**

fetching this url (with a valid token) and using the code from the official source available here
https://gateway.temporal.cloud/ipns/docs.api.temporal.cloud/ipfs.html#get-uploads

will return an ERR_CONTENT_LENGTH_MISMATCH 200 (OK)
(with nevertheless a 200 ok status)

but the Fetch will fail without the possibility to retrieve the blob

**2) As a "random empirical" workaround i saw a part of the official api doc where one could fetch the same endpoint thru a specific port 6767**

https://api.temporal.cloud:6767/v2/ipfs/utils/download/
in the private network session available here

https://gateway.temporal.cloud/ipns/docs.api.temporal.cloud/ipfs.html#post-download-file

it worked during a time but now it is constantly returning the following CORS errors reject:

_"has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled."_

I am pretty sure this time you will be able to reproduce the same bugs

Thanks again for your reactivity and effort in the ipfs nodes

ps: I have never had any problem with the add (upload) endpoint

postables:
#1 is a known issue with the work around being what you're doing in step 2. I thought I updated the docs with this but I must've not pushed the update I'll do that. 

#2 you say 'in the private network session' what do you mean by this? Private networks should never be used with `api.temporal.cloud` since they are unsupported in production.

 Can you please provide sample code to trigger the issue in #2

chisNaN:
THanks again for your reactivity

- Ok for point 1
- For point 2 i mean, in order to find this specific port 6767
I had to go in this path of the documentation
/ **Private** network usage / Post download file

I think this part of the api will need more diagnose because when i woke up this morning
it started to work again for the following browsers:

- chrome and firefox under macOS and Android

Nevertheless

- it works "randomly" with opera
- it has never worked under safari macos
(on safari it always end with a TypeError: connection lost)
**- it has also never worked for any browsers under IOS and this is very annoying IMO**

[...]

PR #366, opened 2019-09-18
cors fix 

postables:
Our CORS middleware was having some issues, in part not correctly handling injection of the CORS header. Closes https://github.com/RTradeLtd/Temporal/issues/365

## :rocket: Changes
* Switches CORS middleware to `rs/cors` and ensure we appropriately whitelist all needed headers, and methods.
* Enables go1.13

## :warning: Breaking Changes
None


go.mod
```diff
@@ -41,7 +41,6 @@ require (
 	github.com/ipfs/go-ipfs v0.4.22 // indirect
 	github.com/ipfs/go-ipfs-addr v0.0.1
 	github.com/ipfs/go-path v0.0.7
-	github.com/ipfs/go-peertaskqueue v0.0.5-0.20190704154349-f09820a0a5b6 // indirect
```
xiegeo: Also at line 32 `github.com/gin-contrib/cors v1.3.0` shouldn't be imported anymore, right?

postables: yea, had to tidy some dependencies.

chisNaN:
Thanks for your reactivity and efforts.
Nevertheless i dont know **if it has to be reopened**
It is stable on chrome, firefox macos and android
**but still not working on safari macos and neither browsers on ios**

I have made this test page on netlify so you can check with any user agent.
https://temporal-cloud-download.netlify.com
You just need to pass a valid token in the url like this 

`/?token=bla`

This page show in html the javascript code behind in case of

:bulb: For the moment i stick with the proxify solution which is working with all browsers

Issue #369, opened 2019-09-23
Enable Organizational Account System


postables:
* Enables creation of "sub-user" accounts
* Enables monitoring of all sub-user accounts
* Enables billing rate proxying to sub-user accounts

postables:
See https://github.com/RTradeLtd/database/pull/46 for work being done

PR #376, opened 2019-10-16
Organization Billing

postables:
In certain situations a more complex billing structure is needed, that may vary depending on the use-case. In order to avoid the business logic many times, and for each specific use-case. This allows creating groups of users called "organizations", and collectively gathering up the total storage, and cost of storage used by all accounts under the organization.

This change effectively side-steps the credit process, and all user accounts that are registered under an organization do not need to interact with the credit system.

Closes https://github.com/RTradeLtd/Temporal/issues/369, requires https://github.com/RTradeLtd/database/pull/46 be merged first.

## :rocket: Changes

* Enables creating organizations
* Enables registering users under the organization
* Organization users get billed at $0.00/gb, with the raw cost of the storage at partner tier being applied as a debt towards the organization
* Updates xss middleware https://github.com/dvwright/xss-mw

## :warning: Breaking Changes

* None!


postables:
There is a bug with org user registration that can be reproduced with the following steps:

* Create organization
* Register org user but specify an organization name that does not exist
* User account creation fails, but the email + username are stored in database.

This means we aren't properly doing input validation.

postables:
It now appears that signatures generated for email validation of org registers users results in an invalid signature

api/v2/routes_organizations_test.go
```diff
@@ -0,0 +1,107 @@
...
+		t.Fatal(err)
+	}
+	defer db.Unscoped().Delete(org)
+
+	// register organization user
+	testRecorder = httptest.NewRecorder()
+	req = httptest.NewRequest("POST", "/v2/org/register/user", nil)
+	req.Header.Add("Authorization", authHeader)
+	urlValues = url.Values{}
+	urlValues.Add("username", "testorg-user")
```
xiegeo: we can add test cases like `urlValues.Add("username", "")` too

PR #399, opened 2019-11-24
Remove Zip Directory Upload

bonedaddy:

Zip file based directory uploads was a feature we added experimentally in dev, but never moved to production. It was never used outside of our development environment, so it's being removed, and considered an "unbreaking change" as this was a possibility since conception of the zip file based directory uploader.

This is now something that can be handled via our iPFS HTTP API reverse proxy.

## :rocket: Changes

* Remove `/v2/add/directory` call
* Remove clamav tooling

## :warning: Breaking Changes

None


bonedaddy:
Nope

Issue #400, opened 2019-11-24
Cache CMC Price Lookups


bonedaddy:
**Why are you submitting this feature request**
Currently we dont cache CMC request looks ups.

**Describe the solution you'd like**
Cache CMC request lookups, only hitting the api every 5 min

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #416, opened 2019-12-11
Cache CMC Price Lookups

bonedaddy:

Closes https://github.com/RTradeLtd/Temporal/issues/400 and enables an in-memory cache of cmc prices to prevent excessive requests.

## :rocket: Changes

* Cache successful cmc price lookups in-memory
* When a cmc price lookup fails, return the in-memory price and error
* Within the API, ignore errors when retrieving usd prices *if* the returned price is non-zero

## :warning: Breaking Changes

None

### DIFF
--- a/utils/cmc.go
+++ b/utils/cmc.go
@@ -50,7 +50,12 @@ func init() {
 	}
 }
 
-// RetrieveUsdPrice is used to retrieve the USD price for a coin from CMC
+// RetrieveUsdPrice is used to retrieve the USD price for a coin from CMC.
+//
+// Whenever we have a "fresh" coin price that is newer than 10 minutes
+// we will return that price instead of querying coinmarketcap. In the event
+// of a "stale" value we will hit the coinmarketcap api. If that errors
+// then we return both the error, and whatever price we have in-memory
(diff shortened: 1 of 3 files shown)


utils/cmc.go
```diff
@@ -31,8 +43,28 @@ type Response struct {
...
+			goto REFRESH
+		}
+		cost := pricer.coins[coin].price
+		pricer.mux.RUnlock()
+		return cost, nil
+	}
+REFRESH:
+	pricer.mux.RUnlock()
+	pricer.mux.Lock()
+	defer pricer.mux.Unlock()
```
xiegeo: Add another freshness check here (while mux is write-locked). If two RetrieveUsdPrice of the same coin was called, they could both trigger REFRESH before nextRefresh gets updated.

xiegeo:
We could also handle temporary lookup errors over the network by reusing the old outdated value when updating the cost returns an error.

utils/cmc.go
bonedaddy: good idea

bonedaddy:
> We could also handle temporary lookup errors over the network by reusing the old outdated value when updating the cost returns an error.

Good idea

PR #419, opened 2019-12-14
Add Better Monitoring Service Detect

bonedaddy:

The previous service monitoring script was brittle and doesn't properly handle all types of invocations.

## :rocket: Changes

Instead of grepping on complete command and filtering, grep the count of command names, so instead of `temporal queue ipfs ipns-entry` we grep on `queue ipfs ipns-entry`

## :warning: Breaking Changes

None


xiegeo:
have you checked out `monit`?

bonedaddy:
Not yet, I don't want to make any changes to the way Temporal is monitored just get since these scripts are used with Zabbix which works quite well

Issue #425, opened 2019-12-24
Enable Usage Of Statticcheck


bonedaddy:
[minio](https://github.com/minio/minio/blob/master/Makefile#L48) is currently using [staticcheck](https://staticcheck.io/) which upon further inspection is pretty useful. We should enable it for Temporal, and ideally all Temporal public repositories.

PR #434, opened 2020-01-11
Org Refactor

bonedaddy:

Pull in organization refactor, and some cleanup. Closes https://github.com/RTradeLtd/Temporal/issues/425

## :rocket: Changes

* Add helper function to invoke post-account creation handler
* Pull in database update to include refactored billing report
* Enable returning a paginated response of org user(s) uploads, or return org user(s) uploads as a CSV file
* Implements staticcheck in ci
* Fixes staticcheck issues

## :warning: Breaking Changes

Technically the billing report response is different, but its an experimental feature so :woman_shrugging:


bonedaddy:
~There's a bug when requesting a non-csv user upload, the response contains nothing, but if using a single user it works.~

Going to scrap returning multiple users at once, and give two options:
* Paginated individual user uploads
* Complete uploads for a single user returned as a CSV file

api/v2/routes_frontend_test.go
```diff
@@ -32,7 +32,7 @@ func Test_API_Routes_Frontend(t *testing.T) {
 	fakeSigner := &mocks.FakeSignerClient{}
 	fakeWalletService := &mocks.FakeWalletServiceClient{}
 
-	api, testRecorder, err := setupAPI(t, fakeLens, fakeOrch, fakeSigner, fakeWalletService, cfg, db)
+	api, _, err := setupAPI(t, fakeLens, fakeOrch, fakeSigner, fakeWalletService, cfg, db)
```
xiegeo: I think the return of testRecorder should be removed from setupAPI. It is never acturally used in any test code.

api/v2/routes_organizations.go
```diff
@@ -167,135 +143,144 @@ func (api *API) registerOrgUser(c *gin.Context) {
...
+		if err != nil {
+			continue
+		}
+		resp[user] = uplds
+	}
+	return resp, nil
+}
+
+// returns true if user is owner
+func (api *API) validateOrgOwner(c *gin.Context, organization, username string) (*models.Organization, bool) {
```
xiegeo: I think returning (*models.Organization, error) is better, where err == nil is ok. so the error can be handled by the caller.

queue/queue.go
```diff
@@ -49,32 +49,36 @@ func New(queue Queue, url string, publish, devMode bool, cfg *config.TemporalCon
...
 		// see https://godoc.org/github.com/streadway/amqp#DialTLS for more information
 		tlsConfig := new(tls.Config)
 		tlsConfig.RootCAs = x509.NewCertPool()
-		ca, err := ioutil.ReadFile(cfg.RabbitMQ.TLSConfig.CACertFile)
+		ca, err = ioutil.ReadFile(cfg.RabbitMQ.TLSConfig.CACertFile)
 		if err != nil {
-			return nil, err
+			break
 		}
 		if ok := tlsConfig.RootCAs.AppendCertsFromPEM(ca); !ok {
```
xiegeo: can be simlified to `if !tlsConfig.RootCAs.AppendCertsFromPEM(ca) {`

queue/queue.go
```diff
@@ -49,32 +49,36 @@ func New(queue Queue, url string, publish, devMode bool, cfg *config.TemporalCon
...
+		cert, err = tls.LoadX509KeyPair(cfg.RabbitMQ.TLSConfig.CertFile, cfg.RabbitMQ.TLSConfig.KeyFile)
 		if err != nil {
-			return nil, err
+			break
 		}
 		tlsConfig.Certificates = append(tlsConfig.Certificates, cert)
 		conn, err = amqp.DialTLS(connectionURL, tlsConfig)
+		if err != nil {
+			break
+		}
```
xiegeo: Can be just `return amqp.DialTLS(connectionURL, tlsConfig)` here. I would also use `return nil, err` instead of break for the other error conditions.

api/v2/routes_organizations.go
bonedaddy: There is no need to handle the error here. The only "handle" is return early and give an error.

api/v2/routes_frontend_test.go
bonedaddy: Good idea

queue/queue.go
bonedaddy: Good idea

queue/queue.go
bonedaddy: good idea

Issue #433, opened 2020-01-05
Enable Paginated Responses From APIs


bonedaddy:
**Why are you submitting this feature request**
Certain APIs can have massive payloads, for example accounts that have thousands of uploads.

**Describe the solution you'd like**
Be able to return paginated results from APIs that return large payloads

**Describe alternatives you've considered**

N/A

**Additional context**

~https://github.com/biezhi/gorm-paginator~ Use our fork https://github.com/RTradeLtd/gpaginator

bonedaddy:
Example
```Go
	// TODO(bonedaddy): add tests
	if c.Param("paged") == "true" {
		page := c.Param("page")
		if page == "" {
			page = "1"
		}
		limit := c.Param("limit")
		if limit == "" {
			limit = "10"
		}
		pageInt, err := strconv.Atoi(page)
		if err != nil {
			Fail(c, err, http.StatusBadRequest)
			return
		}
		limitInt, err := strconv.Atoi(limit)
		if err != nil {
			Fail(c, err, http.StatusBadRequest)
			return
		}
		var uploads []models.Upload
		paged, err := gpaginator.Paging(
			&gpaginator.Param{
				DB:    api.upm.DB.Where("user_name = ?", username),
				Page:  pageInt,
				Limit: limitInt,
			},
			&uploads,
		)
		if err != nil {
			api.LogError(c, err, "failed to get paged user upload")
			return
		}
		Respond(c, http.StatusOK, gin.H{"response": paged})
		return
	}
```

PR #437, opened 2020-01-16
Enable Paginated Responses For Large Payload Calls

bonedaddy:

Closes https://github.com/RTradeLtd/Temporal/issues/433

## :rocket: Changes

The following calls now optional return paginated responses:
* Uploads
* Uploads by network
* Encrypted uploads
* IPNS Records

You can invoke paginated responses using `?paged=true` url parameter. Additionally there are two "optional" parameters to control the response:
* `page` = page number (default 1)
* `limit` = max number of results (default 10)

## :warning: Breaking Changes

None this is backwards compatible. Note however this is an "experimental" feature, and the api is subject to change until finalized. That means although we probably wont change the `paged`, `page` and `limit` url params, it's possible.

### DIFF
--- a/api/v2/api_test.go
+++ b/api/v2/api_test.go
@@ -78,6 +78,43 @@ type stringSliceAPIResponse struct {
 }
 
 // sendRequest is a helper method used to handle sending an api request
+func sendRequestPaged(
+	api *API,
+	method, url string,
+	wantStatus int,
+	body io.Reader,
+	urlValues url.Values,
+	out interface{},
+) error {
+	testRecorder := httptest.NewRecorder()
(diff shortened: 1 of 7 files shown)


api/v2/api_test.go
```diff
@@ -77,6 +77,43 @@ type stringSliceAPIResponse struct {
...
 }
 
+// sendRequest is a helper method used to handle sending an api request
+func sendRequestPaged(
+	api *API,
+	method, url string,
+	wantStatus int,
+	body io.Reader,
+	urlValues url.Values,
+	out interface{},
```
xiegeo: Some documentation for out would be helpful.

Issue #446, opened 2020-02-05
In Dev Mode Key Export Fail


bonedaddy:
**Describe the bug**
When running the API in dev mode, key export will fail. This is because the API expects to delete the keys from two krab keystore backends, and when in dev mode there is only 1 keystore backend. Therefore we need to have a check similar to the queue process where  if in dev mode we only delete from the primary krab keystore and when in prod we delete from both.

This bug allows removal of the key from the keystore, but not the database.

**To Reproduce**
1) Run API in dev
2) Create key
3) Delete key (you will get an error)

**Expected behavior**
Delete key with no error

Issue #447, opened 2020-02-05
Enable Insecure DB Connections For Queue


bonedaddy:
The API allows falling back to insecure DB connections which is useful for dev environments. The queue system currently only allows secure connections. Enable automatic fallback to insecure, and if that fails, exit.

bonedaddy:
Closed

PR #448, opened 2020-02-05
Misc Improvements

bonedaddy:

Closes https://github.com/RTradeLtd/Temporal/issues/446, https://github.com/RTradeLtd/Temporal/issues/447

## :rocket: Changes

See corresponding tickets

## :warning: Breaking Changes

None


bonedaddy:
No, that's a breaking change which would cause existing tooling to break with a flag undefined error from the command line. We should do it when we have a breaking release though

PR #450, opened 2020-02-09
Xiegeo/debug

xiegeo:
Some changes to the code during deployment testing.

@bonedaddy Please choose which changes you want to accept.

## :rocket: Changes
- Workaround for using a newer version of ipfs-cluster.
- Shows the absolute location of expected certs if there was an error reading it.
- A few spelling fixes.

## :warning: Breaking Changes
none

### DIFF
--- a/queue/ipfs_cluster.go
+++ b/queue/ipfs_cluster.go
@@ -73,7 +73,7 @@ func (qm *Manager) processIPFSClusterPin(ctx context.Context, d amqp.Delivery, w
 		return
 	}
 	qm.l.Infow(
-		"pinning has to cluster",
+		"pinning hash to cluster",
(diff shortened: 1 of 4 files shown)


queue/ipfs_cluster.go
```diff
@@ -87,6 +87,9 @@ func (qm *Manager) processIPFSClusterPin(ctx context.Context, d amqp.Delivery, w
 		d.Ack(false)
 		return
 	}
+	if err != nil && err.Error() == "json: Unmarshal(nil) (200)" {
+		qm.l.Infow("compat bug", "error", err)
```
bonedaddy: Change this to an error log not an info log, it will be more likely to be picked up by manual inspection, or filter out issues with log scrapers.

PR #466, opened 2020-04-03
Better Welcome Email

bonedaddy:

When doing third-party implementations (either non-temporal api clients, or via other services like ENS) it's important that during sign up we communicate information about the pinning system. Also our original welcome email was quite frankly a pos lol.

This should also help clear up a lot of the question we get from first time users.

## :rocket: Changes

* Nicer welcome email message explaining the pinning system
* Use an emoji cause it looks cool

## :warning: Breaking Changes

* None


DerrickF86 (approved):
As my first review and new GitHub King I hearby delcare this officially approved.

PR #467, opened 2020-04-04
Increase Free Tier Maximum Pin Time To 12 Months

bonedaddy:

We're adjusting the maximum time that free accounts can pin data for, and removing the limitation where free users are unable to extend their pins. The new maximum limit is 12 months for free users, while paid users stay the same at 24 months. 

The database dependency update will pull in the ability to remove pins + claim partial refunds, however this will be enabled in a different PR

## :rocket: Changes

* Update database dep adding ability to remove pins but not enabling
* Set maximum free tier pin time to 12 months
* Enable free tier users to extend their pins for up to 12 months
* Update welcome email message to indicate that warp (our IPFS HTTP API) endpoint for free users will set the pin time to 12 months, but not do this for  paid tier accounts. We dont want to overcharge paid users, so paid users must as usual extend pins.

## TODO

* Update database dependency once merge
* Update new playground to reflect the change (different PR)
* Update playground to enable removing pins (differnet PR)


api/v2/utils.go
```diff
@@ -255,13 +255,21 @@ func (api *API) validateHoldTime(username, holdTime string) (int64, error) {
 	return holdTimeInt, nil
```
xiegeo, line 255: It doesn't let me comments on the line, but there is an error message that needs to be updated 5 lines above.

api/v2/routes_rtfs.go
```diff
@@ -381,8 +375,11 @@ func (api *API) extendPin(c *gin.Context) {
 		api.LogError(c, err, eh.UploadSearchError)(http.StatusBadRequest)
 		return
 	}
+	// indicate whether or not the user is free
+	// will be true if usg is free tier
+	var isFree = usg.Tier == models.Free
 	// ensure even with pin time extension, it wont breach two year limit
-	if err := api.ensureTwoYearMax(upload, holdTimeInt); err != nil {
+	if err := api.ensureLEMaxPinTime(upload, holdTimeInt, isFree); err != nil {
```
xiegeo, line 382: If the function signature is changed to accept `usg.Tier` directly, it's better for type safety and future-proofing possible changes to max pin time.

utils/cmc.go
```diff
@@ -120,6 +120,9 @@ REFRESH:
 		}
 		if quotemap["quote"] != nil {
 			b, err = json.Marshal(quotemap["quote"])
+			if err != nil {
+				return pricer.coins[coin].price, err
+			}
 			if err := json.Unmarshal(b, &usdmap); err != nil {
 				fmt.Println("error: ", err)
 				return pricer.coins[coin].price, err
```
xiegeo, line 128: Nice catch, but why is it marshalling and unmarshalling the same data? Is it using marshalling to do a deep copy?

bonedaddy: long story short, coinmarketcap absolutely butchered they're API, and the index keys for the JSON they send back is dynamic and entirely depends on the total marketcap of the cryptocurrency. It's also not possible to use golang structs to unmarshal this data, because we can't use JSON tags because the index key for the data we want isn't static. So we need to do this absolutely grotesque method of unmarshaling to maps just to get access to the data we want. 

Unless I had a big brain fart, its one of the most poorly design APIs that sends back JSON responses ever.

bonedaddy:
> I think where possible, we should use time.Duration for all internal duration representations, instead of a mix of months and hour in ints and floats.

We already sort of do this,  "months" are just used to feed into `time.Now().AddDate`; The problem with using `time.Duration` for this is that it doesn't have a concept of months, and the maximum time period is `hour` increments. The only thing we use hours for is the refund calculations, which uses `time.Duration` and `time.Time` to derive the hours from the garbage collect date in the upload model which is stored as `time.Time`.

Because we only allow users to pin in month long increments that cap out at 24 I think its ok because all the numeric types we deal with can correctly swap between each other. 

That being said eventually when we do release the v3 API we should migrate to `time.Duration` not have the int, and floats stuff that we do now.

api/v2/routes_rtfs.go
bonedaddy: smart

api/v2/utils.go
bonedaddy: nice catch

utils/cmc.go
xiegeo: 😂

api/v2/utils.go
```diff
@@ -248,28 +248,29 @@ func (api *API) validateHoldTime(username, holdTime string) (int64, error) {
...
+	default:
+		return errors.New("invalid usage tier")
 	}
 	// get current time
 	now := time.Now()
 	// get future time while factoring for additional hold time
 	then := upload.GarbageCollectDate.AddDate(0, int(holdTime), 0)
 	// get the time difference and ensure the appropriate maximum pin time
-	if then.Sub(now).Hours() > hours {
+	if then.Sub(now).Hours() > (time.Hour * time.Duration(hours)).Hours() {
```
xiegeo, line 273: might be better with
```
maxDuration = 8760 * time.Hour //renamed from hours
...
if then.Sub(now) > maxDuration 
```
Or
```
limit = now.AddDate(1,0,0) //this should work with leap years too
...
if then.After(limit)
```

My idea was to use `time.*` types as much as possible. For example, instead of `holdTime int64`, the function accepts a `date.Time`, that was translated from a user input higher up in the call stack. So, a hold time of 12 months is instead passed in as "April 8th, 2021" or at the end of April if that's more appropriate.

Also, some user input validation and error messages are all over the place. So id we also use the time translation function as a validation function, then we can treat `date.*` values as trust worthy.

bonedaddy: Ah okay I see what you're saying smart.

> My idea was to use time.* types as much as possible. For example, instead of holdTime int64, the function accepts a date.Time, that was translated from a user input higher up in the call stack. So, a hold time of 12 months is instead passed in as "April 8th, 2021" or at the end of April if that's more appropriate.
> Also, some user input validation and error messages are all over the place. So id we also use the time translation function as a validation function, then we can treat date.* values as trust worthy.

Smart as well, this will definitely be something to do with v3 since I think it's a breaking change. I will create an issue to track this.

PR #470, opened 2020-04-05
Experimental Swarm Implementation (2nd Storage Network Support)

bonedaddy:

This enables basic support for a 2nd storage network that can be accessed by Temporal. IT allows basic file uploads to the Swarm storage network, sometimes called "Ethereum Swarm" or "etherswarm". 

We use a basic wrapper [client](https://github.com/RTradeLtd/swampi)  to abstract low-level usage of the swarm network. It also allows high-level data replication by distributing data to two different swarm nodes.

## :rocket: Changes

Allow swarm uploads

## :warning: Breaking Changes

None


api/v2/routes_rtfsp_management.go
```diff
@@ -30,9 +30,11 @@ func (api *API) createIPFSNetwork(c *gin.Context) {
...
 		return
 	}
-	// make sure the name is something other than public
-	if strings.ToLower(networkName) == "public" {
-		Fail(c, errors.New("network name can't be public, or PUBLIC"))
+	switch strings.ToLower(networkName) {
+	case "public", "etherswarm":
+		Fail(c, errors.New("networok name cant be public, PUBLIC, etherswarm, or ETHERSWARM"))
+		return
+	default:
```
xiegeo, line 37: I think this default line is unnecessary.

api/v2/api.go
```diff
@@ -62,7 +63,10 @@ type API struct {
 	queues      queues
 	service     string
 	version     string
-
+	swarm       struct {
+		endpoint1 *swampi.Swampi
+		endpoint2 *swampi.Swampi
+	}
```
xiegeo, line 69: Why is there a hard limit on two endpoints?

xiegeo (approved):

api/v2/routes_rtfsp_management.go
bonedaddy: yea it is

api/v2/api.go
bonedaddy: no real reason. at the moment we only need 2, however I've gone ahead and structure it in such a way that you can use more than 2 endpoints if the config package is adjusted to support it

api/v2/utils.go
```diff
@@ -308,36 +308,33 @@ func (api *API) getCaptchaKey() string {
 // dualSwarmUpload enables uploading data to 2 different swarm nodes at once
 // to provide high-availability of stored data
 func (api *API) dualSwarmUpload(data []byte, isTar bool) (string, error) {
```
xiegeo, line 310: This should be renamed now, since the limit of 2 swarms is removed

api/v2/routes_swarm.go
```diff
@@ -13,7 +13,7 @@ import (
 
 // SwarmUpload is used to upload data to ethereum swarm
 func (api *API) SwarmUpload(c *gin.Context) {
-	if api.swarm.endpoint1 == nil {
+	if api.swarmEndpoints == nil {
 		Fail(c, errors.New("must have at least one non nil swarm client"))
 		return
```
xiegeo: len(...) == 0 is more dependable than testing for nil.

Also, the error message should be updated to "must have at least one ~~non nil~~ swarm client"

api/v2/utils.go
```diff
@@ -308,36 +308,33 @@ func (api *API) getCaptchaKey() string {
...
-		resp1.Body.Close()
-		return "", err
-	}
-	hash1 := string(responseData)
-	resp1.Body.Close()
-	if api.swarm.endpoint2 != nil {
-		resp2, err := api.swarm.endpoint2.Send(swampi.SingleFileUpload, bytes.NewReader(append(data[0:0:0], data...)), map[string][]string{
+	var hashes []string
+	for _, endpoint := range api.swarmEndpoints {
+		resp, err := endpoint.Send(swampi.SingleFileUpload, bytes.NewReader(append(data[0:0:0], data...)), map[string][]string{
```
xiegeo, line 313: I'm not sure why `append(data[0:0:0], data...` isn't just data? I think the extra data copy is unnecessary.

xiegeo (changes requested):

api/v2/utils.go
bonedaddy: Good catch, I thought the bytes reader would drain the data but it doesnt

api/v2/routes_swarm.go
bonedaddy: good idea

xiegeo (approved):
Just a question about if `dualSwarmUpload` should be renamed left.

bonedaddy:
@xiegeo ah shoot I forgot about that, let me make the change

xiegeo (approved):

PR #474, opened 2020-04-13
ReCAPTCHA + Rate Limit Refactor + Deduped Pin Cost Calculation Where Possible

bonedaddy:

Enabling optional usage of reCAPTCHA v3, and refactor rate limiter, require verified emails to login. The release of this PR needs to be timed on several factors:

* ~Update warp to include the same logic~
* ~Update database to set unverified limits to 0~
* ~ENS management app updated~
  * ~This is the main coordination aspect. We need to time a deploy of our update with a deploy of the ens management app~

Closes https://github.com/RTradeLtd/Temporal/issues/476

## :rocket: Changes

* Modify rate limiter to go by clientIP
* Enable callbacks to reCAPTCHA v3 
  * Note this doesn't mean you have to use reCAPTCHA/v3 with Temporal, if you don't want to enable captcha verification, don't supply a value either via the environment variable, or config file (default behavior)
* Require verified emails to sign-in
* When calculating pin cost, first attempt a deduplicated size analysis, falling back to traditional size analysis
  * Deduplicated size cost will only be possible with UnixFS objects, as arbitrary ipld types may not conform to the same specifications, and thus require traditional size calculation

## :warning: Breaking Changes

None


api/middleware/jwt.go
```diff
@@ -48,16 +48,24 @@ func JwtConfigGenerate(jwtKey, realmName string, db *gorm.DB, l *zap.SugaredLogg
...
-			if _, err := userManager.FindByUserName(userId); err != nil {
+			usr, err := userManager.FindByUserName(userId)
+			if err != nil {
 				return false
 			}
-			return true
+			if usr.EmailEnabled && usr.AccountEnabled {
+				return true
+			}
+			return false
```
xiegeo, line 68: `return usr.EmailEnabled && usr.AccountEnabled`

bonedaddy: good catch

